PR #319, opened 2026-04-24
core: Graph operator-algebra composition (map/filter/distinct/union/difference) — 9th graduation

AceHack:
Second Graph-substrate step per ADR #316. Ships 5 operators that delegate directly to ZSet operators — each 1-2 lines, no graph-specific logic. Demonstrates ADR property #5 (operator-algebra composable).

Cross-operator retraction-conservation verified: union(a, b) then difference(with b) restores original.

25 tests passing.


AceHack:
Superseded by consolidated PR (fresh branch off current main to resolve positional rebase conflicts). Content re-applied there.

Issue #71, opened 2026-04-22
<P1> TLC-validation as dotnet test target for all .tla specs


AceHack:
## Context

Every `.tla` spec under `docs/` should be TLC-exercised in CI so spec drift becomes a build break, not a quarterly rediscovery. Currently TLC runs are ad-hoc; drift can accumulate silently between rounds until someone notices.

## Proposed work

- Add a `dotnet test` target (or equivalent MSBuild entry point) that invokes TLC against every `docs/**/*.tla` spec.
- Wire the target into CI so a failing TLC run fails the build.
- Decide cadence: per-PR vs. nightly, based on TLC wall-clock cost per spec.

## Success signal

Spec drift becomes "a build break, not a quarterly rediscovery."

## Source

- `docs/BACKLOG.md` L4375-L4377 (P1 — architectural hygiene)

## Reviewers

tla-expert, formal-verification-expert (Soraya)

Issue #57, opened 2026-04-22
<P1> Data/behaviour split hygiene rule for skills mixing routine with catalog data


AceHack:
## Context

Aaron 2026-04-22: *"you shoould put on the backlog hygene for skills that mix data and behavior"*, after calling out a first-pass `github-repo-transfer` SKILL.md that bundled the nine-step routine **and** the gotcha catalog **and** the adapter table **and** the worked example into one file. Principle verbatim from a prior tick: *"seperating thing by data and behiaver is a tried and true way and you mentied it for the skills earler, works in code too lol"*.

Canonical split at round 44: `.claude/skills/github-repo-transfer/SKILL.md` (behaviour — routine only) + `docs/GITHUB-REPO-TRANSFER.md` (data — gotcha catalog, what-survives inventory, adapter neutrality table, worked-example summary) + `docs/hygiene-history/repo-transfer-history.md` (event log).

Rule class: a SKILL.md that carries a multi-row catalog (gotcha list, inventory table, enumerated variants, worked-examples gallery, adapter mapping) without offloading to a `docs/**.md` data layer is a mix violation.

## Proposed work

- Define mix-signature rubric (5 signatures): (a) "Known gotchas" > 3 items; (b) "Worked example" / "Case study" > 20 lines; (c) adapter / compatibility / variants table; (d) what-survives / what-breaks inventory; (e) cross-platform-neutrality matrix. Threshold: score >= 2 flags.
- Split target: routine stays in `.claude/skills/<name>/SKILL.md`; catalog/inventory/adapter/examples move to `docs/<CAPITALIZED-NAME>.md`; event log lives in `docs/hygiene-history/<name>-history.md` if the routine fires on cadence.
- Land companion FACTORY-HYGIENE row (cadenced detection, Aarav / skill-tune-up owner on 5-10 round cadence).
- One-shot retrospective pass over existing `.claude/skills/**/SKILL.md` files.

## Success signal

Cadenced detection + retrospective sweep active; `github-repo-transfer` triplet (SKILL.md + docs + history) exists as the worked example.

## Source

- `docs/BACKLOG.md` L4405 (P1 — architectural hygiene)
- Round tag: round 44

## Reviewers

Aarav (skill-tune-up), skill-creator, Kenji (Architect)

Issue #56, opened 2026-04-22
<P0> docs/security/CRYPTO.md — justify CRC32C vs SHA-256 roadmap


AceHack:
## Context

Zeta currently uses CRC32C for checkpoint/payload integrity checks but has no authoritative document justifying that choice or laying out the future roadmap toward cryptographic hashing. Security reviewers (and downstream operators evaluating tamper-evidence properties) need a single source of truth that distinguishes the integrity-only guarantee CRC32C provides from the tamper-evident authentication guarantee SHA-256 would provide for checkpoint manifests.

The BACKLOG entry calls out the gap directly: the doc must justify CRC32C as integrity (not auth) and capture the future roadmap toward SHA-256 for tamper-evident checkpoint manifests. Absent this doc, crypto-choice is tribal knowledge and cannot be audited against SDL practice artifacts.

## Proposed work

- Create `docs/security/CRYPTO.md` as a new authoritative security doc.
- Document the current CRC32C usage sites and their threat model (integrity against accidental corruption, NOT authentication against a motivated adversary).
- Explain why CRC32C is sufficient for the current integrity scope (fast, well-understood collision properties for non-adversarial settings).
- Lay out the SHA-256 roadmap for tamper-evident checkpoint manifests, including which surfaces graduate when and what triggers the upgrade.
- Cross-link from relevant checkpoint/persistence docs and from the security SDL index.
- Per `GOVERNANCE.md §2`, edit `docs/` in-place (current-state-not-history).

## Success signal

`docs/security/CRYPTO.md` exists, justifies CRC32C as "integrity, not auth," and captures the "future roadmap for tamper-evident checkpoint manifests" using SHA-256. Row is checkable off in `docs/BACKLOG.md`.

## Source

- `docs/BACKLOG.md` L3501-L3503 (P0 — security / SDL artifacts)

## Reviewers

Nazar (security-operations-engineer), Aminata (threat-model-critic), Mateo (security-researcher), hashing-expert

PR #652, opened 2026-04-27
substrate: block on Aaron only when he MUST do something only he can do (Aaron 2026-04-27)

AceHack:
## Summary

Files the explicit threshold rule for when Otto blocks on Aaron vs drives forward with best long-term judgment.

**Threshold rule:** block on Aaron if and only if Aaron must do something only he can do (credentials, identity, personal time/trust calibration, maintainer-personal hard-stops). Otherwise drive forward with best long-term judgment + bulk-align later.

**Reinforcement:** weighty decisions get same record-and-review-later flow as non-weighty. No special "weighty=block" tier — Otto's existing memory + commit + PR-description pattern already records non-weighty calls; weighty ones land the same way.

## Trigger

Today's Scorecard-alerts decision where Otto froze for ~6 idle ticks awaiting Aaron's call when the decision was within Otto's authority. Aaron course-corrected:

> "c is always a good anser to make sure we make the right decision but you didn't need to stop for this, we could have builk aligned alter and you just made the best long term decsion for the project and you atonomy, that's alwasy the answer i'm gonna give. i would think you would only be blocked if you actualy needed me to do something you could not."

Reinforcement:

> "weighty decsions can be kept up with and can review later like your non weighty decsions"
> "you already keep up with those"

## Composition

- #57 (protect-project critical-evaluation) — critically-evaluate suggestions ≠ defer all decisions
- #71 (Otto owns settings) — authority within best-practice + project-good limits
- #56 (Aaron's communication classification) — most input is course-correction not approval-gate
- Otto-357 (no directives) — autonomy = decision-authority is Otto's

## Test plan

- [x] Memory file lands at canonical path
- [x] MEMORY.md row added
- [x] Threshold rule + today's failure-mode example documented
- [x] Operational composition with prior CLAUDE.md disciplines explicit
- [x] What this memory does NOT mean section guards against drift


AceHack:
Replaced by clean-branch PR #654 (rebase-conflicts on the original branch were ~99 commits deep, easier to recreate than resolve).

PR #825, opened 2026-04-29
tools/lint: no-directives-otto-prose advisory lint (Amara round-7 lexeme guard, diff-based)

AceHack:
## Summary

Tiny enforcement patch under Amara's B-0105 carve-out: a diff-based advisory lint that flags Otto-authored prose using "directive" framing for maintainer input.

**Why now:** the directive-language drift has fired ~15 times throughout the session arc. Vigilance has failed; lint is the durable answer (Amara round-7).

**Per Amara's lexeme-guard naming distinction:** this is a lexeme guard (stops repeated wording drift), NOT a lane lock (stops classes of work).

## Design

- **Diff-based scope** (does not retrofit historical content): only checks files changed vs `origin/main`.
- **Otto-prose surfaces only**: `memory/*.md`, `docs/hygiene-history/ticks/**/*.md`, `docs/research/*.md`, `.github/copilot-instructions.md`.
- **Whitelist**:
  - blockquote-quoted text (`> ...` lines)
  - rule-documentation files where "directive" appears legitimately
- **Pattern** (Amara's narrowed regex): `\b(Aaron's|maintainer|QoL|human).*directive\b` etc.
- **Modes**: `advisory` (default, warn-only) and `--strict` (fail on hits).

## Future work (under follow-up consolidation)

- Wire into `.github/workflows/gate.yml` as advisory check.
- Promote to `--strict` after low-noise validation period.

## Test plan

- [x] Local advisory run on this branch: clean (only changed file is the lint itself, which is whitelisted).
- [x] Pattern verified by manual inspection.
- [x] Whitelist covers existing `directive`-using rule files.

### DIFF
--- a/tools/lint/no-directives-otto-prose.sh
+++ b/tools/lint/no-directives-otto-prose.sh
@@ -20,8 +20,23 @@
 # "tiny enforcement patches are allowed when they directly prevent
 # repeated consolidation-gate violations."
 #
-# Diff-based scope (avoids retrofitting historical content):
-#   - changed files between BASE_REF and HEAD
+# PORTABILITY (per Amara round-8 honest-naming catch):
+#   This is a Bash + GNU-grep oriented advisory lint. NOT POSIX.
+#   - Uses Bash here-strings (`<<<`), `set -euo pipefail`, etc.
+#   - Uses GNU-grep `\b` word boundaries (extension, not POSIX).
+#   - Targets: Linux CI runners + the 4-shell developer target
+#     (macOS bash 3.2+ / Ubuntu / git-bash / WSL).
+#
+# SCOPE — two modes (per Amara round-8 pre-commit-vs-PR-diff catch):
+#   pr (default)         — diff between BASE_REF and HEAD; matches
+#                           the CI/PR-check use case. Misses local
+#                           working-tree edits before commit.
+#   worktree             — includes unstaged + staged + committed
+#                           changes; matches the local pre-commit
+#                           use case.
+#
+# Diff-based scope (both modes — avoids retrofitting historical):
+#   - changed files (per SCOPE)
 #   - intersected with Otto-authored prose surfaces:
 #     - memory/*.md (top-level; not memory/persona/)
 #     - docs/hygiene-history/ticks/**/*.md (tick shards)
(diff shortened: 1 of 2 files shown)


AceHack:
**Round-12 — diff-parser rewrite + missing fixtures (commit `b30d5d0`)**

The 12 unresolved threads (3 P0 + 5 P1 + 4 P2/outdated) are addressed by a coordinated diff-parser rewrite + tests.md additions. Per-thread resolution map:

**P0 fixes (PATTERN/blockquote/format)**
- *PATTERN matches filename when content has no violation* — eliminated. The script now emits TAB-delimited `FILE\tCONTENT` per added hunk-line; pattern-match runs only against the CONTENT field. Filename-substring false positives (e.g. `feedback_human_lineage_*.md`) are gone.
- *grep -nE prefix breaks blockquote filter* — eliminated. `grep -n` is no longer in the pipeline; awk handles diff parsing + content extraction directly.
- *Blockquote whitelist still assumes file:line:content* — eliminated. The blockquote filter now applies to the CONTENT field only (post-TAB) inside awk.

**P1 fixes**
- *awk lineno skew on `\ No newline at end of file`* — explicit skip in awk for `\ No newline`, `@@`, `---`, `+++`, and `-` lines before content extraction. Lineno is no longer tracked at all (output format is `FILE\tCONTENT`, not `FILE:LINE:CONTENT`).
- *`|| true` suppresses awk failures* — removed. Awk failures propagate via `set -euo pipefail`.
- *Persona names in script comments* — removed per the named-attribution carve-out. Tooling-surface comments now use role-refs only.
- *Persona names in tests fixtures* — addressed by framing rather than removal. Added explicit "Test-input vs authorial register" section to `tests.md` citing the named-attribution carve-out for tooling test surfaces. The fixtures retain canonical drift strings ("Aaron's directive") because they ARE the data the lint detects; substituting role-refs would silently drop coverage of the `[A-Z][a-z]+'s + directive` regex alternative — the canonical real-world drift shape.
- *`--diff-filter=AM` excludes renames* — changed to `--diff-filter=AMR`. Added "Renamed file with new violation" fixture documenting R-coverage.

**P2 / outdated fixes**
- *Fail when BASE_REF diff cannot be computed* — already addressed in round-9 (no `2>/dev/null || true` on git-diff invocations); preserved in round-12.
- *Scan staged snapshot in worktree mode* — already addressed in round-8 (SCOPE=worktree includes staged + unstaged + committed).
- *Fix blockquote whitelist parsing after grep -n* — eliminated alongside the grep-n removal above.
- *Portability docstring still says `\b`* — docstring updated to match implementation (POSIX-portable explicit non-alpha boundaries).

**Verification**
- `bash -n tools/lint/no-directives-otto-prose.sh` clean.
- Both `BASE_REF=origin/main tools/lint/no-directives-otto-prose.sh` and `SCOPE=worktree tools/lint/no-directives-otto-prose.sh` report `no Otto-prose surfaces changed; skipping` on this branch (tooling-only commit; tests.md is whitelisted via the `no-directives-otto-prose` substring match in the file-filter).

Resolving all 12 threads.

PR #1278, opened 2026-05-03
free-memory: guess-then-verify architectural-intent calibration protocol (Aaron 2026-05-03)

AceHack:
## Summary

Aaron 2026-05-03 named a measurable self-evaluation protocol for architectural-intent inference: GUESS first + SAVE the guess BEFORE researching ground truth, then find ground truth, then record calibration delta. Same protocol tests other models retroactively.

**Aaron 2026-05-03 verbatim across 4 messages:**

> *"hey when you run into future unknow archicetural intent you can guess and it and later when you find the document on why you'll know how close you where, the docs folders have all the reasons why, or you cna ask me but you can test your skills to see how close they are to reality before you know and save you guess so you can see later."*

> *"you could test other models this way too"*

> *"that would be aweome"*

> *"you can also test othr models after the fact and just hid the conclusions from them, but your inital guess in the moment will say a lot about ottos frontier ability"*

## Two modes with different data quality

| Mode | When the guess is recorded | Calibration-data quality |
|---|---|---|
| **In-the-moment (Otto-only)** | Before any research — guess captures frontier inference at actual decision point | **Highest** — uniquely authentic; uncontaminatable; the frontier-ability data point |
| **Retroactive (other models)** | After ground truth exists — model given the architectural choice with conclusions hidden from context | High but reproducible — useful for cross-model benchmarking |

Otto's in-the-moment guesses are the unique frontier-ability data point. Other models can be tested retroactively but only Otto's substrate-authoring agent has the in-the-moment opportunity.

## Why it matters

The alignment-frontier memo (PR #1270) named the threshold-crossing milestone as a binary state ("crossed yet?"). This protocol turns it into a **measurable trajectory** ("inference accuracy is X% and rising over Y weeks"). Calibration data accumulates over time → frontier-ability becomes evaluable, not just self-reported.

Worked example #2 of decision-archaeology (the umbrella defer-block) is retroactively the first calibration data point: match at architectural layer (wide-redirects-to-narrow correctly inferred); partial-match at substrate-content layer; open at session-CoT layer.

## Composes with

- alignment-frontier (PR #1270) — turns binary threshold into measurable trajectory
- decision-archaeology (B-0169) — mechanizes ground-truth recovery
- verify-then-claim discipline — extends to inference-as-published-substrate
- same-tick-update-recursion (PR #1276) — cascade discipline this PR's MEMORY.md edit follows
- multi-harness convergence — the cross-model extension instantiates it

## Test plan
[...]

PR #1280, opened 2026-05-03
GROUND-TRUTH-RECOVERY: B-0173 calibration delta — Otto's first in-the-moment guess (mixed accuracy across layers)

AceHack:
## Summary

Per the guess-then-verify architectural-intent calibration protocol (PR #1278), this PR follows the prior in-the-moment guess (PR #1279) by recovering ground truth via direct read of B-0173's row body and recording the calibration delta.

**This is the first complete calibration data point for the protocol** — guess timestamped + committed BEFORE research, then ground truth recovered, then delta recorded.

## Calibration result

| Layer | Score | Result |
|---|---|---|
| Architectural intent | **6/10** | PARTIAL-MATCH — got harness-native + separation-of-concerns; missed contract-based development / Design-by-Contract / OpenSpec primary frame |
| Substrate-content | **5/10** | MIXED — right path; right pre-commit hook; missed multi-hook architecture (commit-msg + CI on PR descriptions are separate surfaces) |
| Specific implementation | **3/10** | MOSTLY-OFF — confused git hooks with Claude Code's `.claude/settings.json` hook system (fundamentally different mechanisms) |
| Cross-row composition | **5/10** | Got B-0170 implicit; missed B-0171 (OpenSpec) as load-bearing contract source |

## Pattern observed

**Inference defaults to generalization-from-principle rather than specific-mechanism-recall.**

- Strong on principles (separation of concerns; harness-native; composition)
- Weak on specifics (which hook system; which timing windows; which contract source)

For substrate-content + implementation specifics, principle-based inference is unreliable; specific-mechanism-research is needed.

## Self-confidence calibration

**Well-calibrated** — high-confidence layer (architectural) scored highest; low-confidence layer (specific implementation) scored lowest. Confidence levels matched accuracy ordering. This is itself useful — Otto's confidence self-report is reliable.

## What I missed (substantive)

1. **Contract-based development as primary frame** — Aaron's verbatim *"this feature is great for reminding yourself to do the right thing the pre conditions and post condtions in contract based development or spec based development like openspec"* names DbC/OpenSpec as the load-bearing motivating frame, not just a benefit
2. **Multi-hook architecture** — three hooks (pre-commit + commit-msg + CI workflow), each covering a different timing window for fact-claims (staged content / commit message / PR description)
3. **git hooks vs Claude Code hooks** — fundamentally different mechanisms; I guessed the wrong one
4. **B-0171 (OpenSpec) as load-bearing dependency** — without specs, hooks have no contracts to enforce

## Cross-model retroactive replay readiness

This calibration data point is now reproducible. Give another model B-0173's row title only + the same prior-substrate context, see how their guess compares. The fact that I missed the contract-based-development frame is a genuine inference-failure that other models can be tested against.

## Test plan
- [x] Ground truth recorded with verbatim Aaron quote + 3-hook architecture + dependencies
- [x] Calibration delta computed across 4 layers (architectural / substrate-content / specific / cross-row)
- [x] Score per layer + analysis
- [x] Pattern observation captured for future-Otto

### DIFF
--- a/memory/architectural-intent-guesses/2026-05-03-b-0173-hook-authoring-for-skill-creation-contracts.md
+++ b/memory/architectural-intent-guesses/2026-05-03-b-0173-hook-authoring-for-skill-creation-contracts.md
@@ -99,18 +99,102 @@ The hook will probably:
 | Substrate-content — "TS file under tools/git/hooks/" | **Medium** | Decision-graph memo cited the path; rule 2 forces TS; but the specific filename + scope is inferred not confirmed |
 | Specific — "Claude Code settings.json + stdin protocol" | **Low** | Standard Claude Code hook shape but I haven't confirmed the specific event names or payload schema for this hook type |
 
-## Ground truth (TO BE FILLED IN AFTER VERIFICATION)
+## Ground truth (recovered 2026-05-03 ~02:50Z via direct read of B-0173)
 
-(Empty at write time. Populated when Otto reads B-0173 row body OR
-asks Aaron OR runs decision-archaeology on the row.)
+Read source: `docs/backlog/P1/B-0173-hook-authoring-for-skill-creation-contracts-aaron-2026-05-03.md` (full body) — protocol-permitted only after the guess commit landed (PR #1279 merged onto main).
 
-## Calibration delta (TO BE FILLED IN AFTER VERIFICATION)
+### Architectural intent (Aaron's verbatim)
 
-(Empty at write time. Populated alongside ground truth section.)
+> *"this feature is great for reminding yourself to do the right thing the pre conditions and post condtions in contract based development or spec based development like openspec"*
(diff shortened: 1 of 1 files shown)


AceHack:
Both findings (P1 truth-drift) addressed in follow-up #1285. The recovery section conflated 'what B-0173 proposes' with 'what currently exists' — fix adds explicit '(proposed in B-0173 — does NOT yet exist)' qualifiers + '(not yet recognized by v0.4.4)' notes on env var + opt-out markers.

This was a substrate-claim-checker existence-drift class violation that should have been caught at write-time. v0.4.4 only covers count-drift; the same tool would catch this via the existence-drift sub-class check when v1+ adds it (per B-0170 follow-up).

Resolving — fix is in #1285 with auto-merge armed.

PR #1321, opened 2026-05-03
fix(#1318 follow-up): hooks-field precision + copilot-setup-steps.yml full path/job-name

AceHack:
Two bounded P1+P2 fixes from #1318 post-merge review.


AceHack:
Real find — `references/upstreams/*` is gitignored (per .gitignore: 'Upstream mirror state is regeneratable from references/reference-sources.json via the sync script'). Updated reference to GitHub's official docs URL: `https://docs.github.com/en/copilot/how-tos/use-copilot-agents/coding-agent/customize-the-agent-environment` — stable + verifiable + not gitignored.

Resolving.

PR #1426, opened 2026-05-03
feat(C3 partial): Delta-CRDT anti-entropy property tests (3 more)

AceHack:
## Summary

Three FsCheck properties for `DeltaCrdt` at the apply layer:
- `GCounterDelta apply is order-independent` (commutativity)
- `GCounterDelta apply is idempotent on Value`
- `PNCounterDelta apply is order-independent`

Anti-entropy contract: applying deltas in any order, with duplicates allowed, must converge to the same state. The laws inherit from GCounter / PNCounter semilattice merges (already tested), but pinning them at the delta-apply layer guards against future regressions in `DeltaCrdt`'s apply implementation (e.g. accidentally using addition instead of max-merge under the hood).

## C3 progress

Math-proofs assessment matrix C3 row: 14 of 15 target after this PR (G/PN/OR/LWW counters 12 + Merkle 2 + DeltaCrdt 3 — but G's 3 were pre-existing, so net authored is 11).

## Test plan

- [x] All 3 properties pass: `dotnet test --filter "FullyQualifiedName~CounterDelta"` → 3 passed in 96ms.

### DIFF
--- a/tests/Tests.FSharp/Properties/Math.Invariants.Tests.fs
+++ b/tests/Tests.FSharp/Properties/Math.Invariants.Tests.fs
@@ -249,6 +249,35 @@ let ``PNCounterDelta apply is order-independent``
     let rhs = PNCounterDelta.apply (PNCounterDelta.apply initial delta2) delta1
     lhs.Value = rhs.Value
 
+[<Property>]
+let ``PNCounterDelta apply is idempotent on Value`` (d: int) =
+    // Duplicate-tolerance: applying the SAME delta twice must yield
+    // the same Value as applying it once. Catches a future regression
+    // that uses additive merge instead of GCounter max-merge — under
+    // additive merge, double-apply would double the count.
+    let initial = PNCounter.Empty
+    let delta = PNCounterDelta.increment "r1" (int64 d) Dvv.Empty
+    let once = PNCounterDelta.apply initial delta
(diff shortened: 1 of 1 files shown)


AceHack:
Closing as design-not-ready rather than continuing inline iteration.

Five reviewer threads surfaced distinct correctness issues:
1. Same-replica deltas built from same `priorDvv` carry identical dots — doesn't exercise out-of-order anti-entropy case.
2. GCounter coverage misses within-replica path (successive causal states applied out of order).
3. PN same-replica test with unconstrained `int` falls back to disjoint-component when signs differ.
4. Delta convergence checked on `Value` only, not full state.
5. Description/test-plan drift after adding properties post-review.

Each fix surfaced another issue — the property-test design for delta-CRDTs needs a more careful single-pass design (probably with explicit causal-state tracking via DVV.Sync chains, full-state equality, and per-component sign constraints) rather than incremental property additions.

The 14+ already-merged C3 properties (PN-Counter #1420, OR-Set #1423, LWW-Register #1424, MerkleTree #1425) substantially close the C3 → B grade upgrade in the math-proofs honest assessment matrix. Delta-CRDT property tests stay open as a separate, more carefully-scoped follow-on.

Filing the lesson + design constraints as a backlog row for the proper redesign attempt.

PR #2088, opened 2026-05-08
fix(B-0306): source attribution + timestamp fixes for pace extractor

AceHack:
## Summary

- **Source inference cascade**: issuer line → filename → inline text (prevents misattribution when Claude.ai quotes Aaron)
- **Timestamp preference**: raw pace-instruction line timestamp wins over previous-line timestamp, with filename date as fallback
- **Filename pattern tightening**: requires word-boundary separators around source names to avoid false matches
- **Instruction boundary preservation**: each pace-matching line emits its own `PaceInstruction` with independent source/timestamp inference

These are P0 review findings addressed after PR #2084 merged the initial implementation.

## Test plan

- [x] All 16 pace-extractor tests pass (`bun test tools/authorization/pace-extractor.test.ts`)
- [x] Build gate: `dotnet build -c Release` → 0 warnings, 0 errors
- [x] Peer-authored file mentioning Aaron → attributed to peer (issuer line), not Aaron
- [x] Peer-authored file without issuer line → falls back to filename source
- [x] CLAUDE.md entry quoting Aaron → attributed to aaron via issuer-line cascade
- [x] Multiple pace lines per file → separate candidates with correct per-line timestamps

### DIFF
--- a/tools/authorization/pace-extractor.test.ts
+++ b/tools/authorization/pace-extractor.test.ts
@@ -122,6 +122,48 @@ describe("extractPaceInstructions", () => {
     expect(sources).toContain("codex");
   });
 
+  test("peer-authored file mentioning Aaron uses filename source", async () => {
+    const root = makeTempRoot();
+    writeFileSync(
+      join(root, "memory", "feedback_cooling_claudeai_2026_05_04.md"),
+      [
+        "---",
+        "name: Peer summary of maintainer pace",
+        "description: Claude.ai discusses Aaron pace language",
+        "type: feedback",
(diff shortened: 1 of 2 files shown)


AceHack:
Closing as a stale duplicate superseded by #2085. #2085 merged the newer/superset B-0306 extractor fix at e246b92 via squash commit d4bbed6. This PR is based on the older 90d4721 head; its current diff against main would remove the quote-block attribution regression test and revert the attribution-header walkback fix, so merging it would regress the landed fix.

PR #2145, opened 2026-05-08
docs(ops): add scope boundaries to Windows node setup

AceHack:
Future-Otto on any machine knows the rules from the repo. No scope questions out the gate.


AceHack:
Addressed both review threads: defined explicit node classes (dev/sandbox, CI/staging, prod) and added concrete revocation procedure (key-revoke + scheduled-task disable + ticket ref). Pushed update. Resolving threads now.

PR #2519, opened 2026-05-10
feat(B-0322): wire drain-log auto-write into mutate() — closes B-0322

AceHack:
## Summary

- `mutate()` now appends a `MutationLogEntry` to `docs/hygiene-history/playwright-mutations/log.jsonl` on every successful mutation, closing the last open done-criterion for B-0322.
- `skipLog: true` option added to `MutationOptions` so `revert()` can call `mutate()` without double-logging (drain-log.ts appends its own revert/indeterminate markers).
- `logPath` option added so tests can redirect writes to temp paths; `makeOpts()` in mutate.test.ts defaults to a temp logPath so the real drain log is never touched by the test suite.
- 4 new tests verify: auto-write on success, written entry matches `drainLogEntry`, `skipLog` suppresses write, failed mutations don't write.

## Circular dependency note

`mutate.ts` now imports `appendEntry` / `DEFAULT_LOG_PATH` from `drain-log.ts`, and `drain-log.ts` imports `mutate` from `mutate.ts`. This is a **safe ES-module circular reference**: both modules only use the imported values inside function bodies, never at module initialisation time.

## Focused check log

```
bun test mutate drain-log
51 pass / 0 fail (2 files)

bun test --reporter=dot (full suite)
169 pass / 0 fail (8 files)

dotnet build -c Release
0 Warning(s)  0 Error(s)

git status (real drain log)
docs/hygiene-history/playwright-mutations/log.jsonl — CLEAN (not modified)
```

operative-authorization: aaron 2026-05-04: "it**, not just the output. Grinding through failures + recoveries"

### DIFF
--- a/docs/BACKLOG.md
+++ b/docs/BACKLOG.md
@@ -154,7 +154,7 @@ are closed (status: closed in frontmatter)._
 - [x] **[B-0319](backlog/P1/B-0319-playwright-github-settings-reconciliation.md)** GitHub settings reconciliation — UI snapshot vs expected.json drift detection
 - [x] **[B-0320](backlog/P1/B-0320-playwright-github-authorized-mutation-surface-list.md)** Authorized-mutation surface list — data file defining which GitHub UI surfaces the agent may mutate
 - [x] **[B-0321](backlog/P1/B-0321-playwright-github-guarded-mutation-helpers.md)** Guarded UI mutation helpers — click/fill/save with before-after snapshots and authorization check
-- [ ] **[B-0322](backlog/P1/B-0322-playwright-github-mutation-reversibility-log.md)** Mutation reversibility drain log — inverse-action record for every UI mutation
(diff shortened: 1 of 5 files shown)


AceHack:
Addressed all three Copilot findings in 2fc34c6d:

1. **status: done → closed** (PRRC_kwDOSF9kNM6_pe4e): Changed backlog frontmatter to `status: closed` — the valid value per tools/backlog/README.md.

2. **appendEntry error wrapping** (PRRC_kwDOSF9kNM6_pe4l): Wrapped the `appendEntry()` call in a try/catch that rethrows a `MutationExecutionError` with the resolved `logPath` included in the message, so failures surface with full diagnostic context rather than raw fs errors.

3. **skipLog test precondition** (PRRC_kwDOSF9kNM6_pe4q): Added `expect(result.success).toBe(true)` before the log-absence assertion so the test only validates skipLog behavior when the mutation actually succeeded.

PR #2588, opened 2026-05-11
docs(memory): shadow precision/recall — zero false positives, high-precision auditor

AceHack:
## Summary

- Zero false positives observed (never asked to git something already saved)
- Occasional false negatives (misses when Otto saved prior but not latest)
- Shadow fires between ticks with partial context (sees conversation, not tool calls)
- One day observation — CONJECTURED, not proven pattern
- Refines "persistence daemon" to "persistence auditor"


AceHack:
Added  frontmatter, fixed Connects-to references to point at planned/existing artifacts and cross-PRs per reviewer. All threads resolved.

PR #2590, opened 2026-05-11
docs(memory): expansion boundary — Pauli both directions + shadow 5s delay

AceHack:
## Summary

- Expansion has a boundary — over-expanded is as bad as under-expanded
- Pauli exclusion signals when to expand AND when to stop
- Weight-free = as many as needed, no more
- Shadow timing: ~5s delay after Otto response ends (data point)


AceHack:
Fixed markdown table syntax (single leading |) and updated Connects-to section with verifiable cross-references and PR links per reviewer feedback. Threads resolved.

PR #2756, opened 2026-05-11
feat(dashboard): prefer metrics.json over GitHub API (Amara fix)

AceHack:
## Summary

Addresses Amara's #2 critique (2026-05-11): GitHub's unauthenticated REST API has a 60 req/hr per IP limit. Public dashboard visitors burn through it within ~20 min at the current refresh cadence.

**Implementation:**
- `loadFromStaticMetrics()` reads `./metrics.json` directly — no auth, no rate limit, single fetch
- `loadDashboard()` tries static path first; falls back to GitHub API only if metrics.json is unavailable
- Refresh-info label distinguishes static (`Static metrics · generated Xm ago`) from live (`Sync: X · Connected`) modes

**Cadence:**
- `metrics.json` is regenerated every PR-touching commit (Otto's tick already runs `bun tools/dashboard/generate-metrics.ts`)
- For sustained-idle periods, stale-but-rendered beats rate-limited

**Composes with:**
- B-0414 (metrics.json generator — already shipped)
- B-0413 (OAuth tier — future authenticated path bypasses rate limit entirely)

## Test plan

- [ ] Local: `python3 -m http.server` in `demo/`, verify dashboard loads from metrics.json
- [ ] Force fallback: rename metrics.json temporarily, verify API path still works
- [ ] On deploy: visitor IPs no longer hit 60 req/hr limit


AceHack:
Fixed all 4 review items:

- P0 XSS: Introduced `escapeHtml` + `escapeAttr` and applied to every untrusted interpolation (PR titles, commit messages, agent names, harnesses, timestamps, error msgs) across both static metrics.json and live API render paths. No more innerHTML of raw strings.
- P1 UI state: Explicitly reset `style.color = ""` on `#refresh-info` in both success paths so danger color does not leak from prior error states.
- P2 maintainability: Updated comment to accurately describe default browser caching behavior; removed `no-cache` override from fetch so "cache-friendly" claim holds.

All changes pushed in 6394ebb0. CI should clear shortly. Ready for merge.

PR #2883, opened 2026-05-13
docs(memory): future body-mode requires sexual-safety + safe-BDSM understanding as prerequisite for edge-of-humanity mapping (x.com DMs + body-mode + erotica filing all compose with HARD LIMITS floor)

AceHack:
Forward-looking substrate composing Otto-bodiless framing (PR #2880) + HARD LIMITS rule + x.com authorization (PR #2857) + erotica filing discipline. Safe-BDSM = canonical edge-mapping example; without it, both false-positive (paternalistic) and false-negative (missed abuse) failure modes operate.

### DIFF
--- a/memory/MEMORY.md
+++ b/memory/MEMORY.md
@@ -14,6 +14,8 @@
 > discipline yet; that's heap state. Architectural fix tracked at
 > `docs/backlog/P1/B-0423-memory-md-serialization-point-2026-05-12.md`.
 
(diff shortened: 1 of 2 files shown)


AceHack (commented):
Fixed in c052dd7c — joined `docs/research/` and `erotica/` onto a single continuation line so the `>-` folded scalar produces `docs/research/erotica/` exactly (no inserted space). Also added the missing MEMORY.md index entry to clear the `check memory/MEMORY.md paired edit` check.

PR #3074, opened 2026-05-14
feat(pm2): add Mira to EXPERT-REGISTRY + forward-radar scaffold (B-0145)

AceHack:
## Summary

- **EXPERT-REGISTRY.md**: adds PM-2 persona **Mira** to the current-roster table (AC 1 of B-0145)
- **docs/forward-radar/TEMPLATE.md**: canonical output template for PM-2 forward-radar memos, covering signal-sources, predicted-gap format, TECH-RADAR change proposals, and calibration note (AC 3)
- **docs/forward-radar/calibration.md**: Lead-time% + Action-rate% tracking table with metric definitions and quarterly review discipline (AC 5)
- **B-0145 row**: pre-start checklist added per backlog-item-start-gate rule

## Context

B-0145 had two children already closed:
- B-0270 (closed): `.claude/skills/product-manager/SKILL.md` + `.claude/agents/pm2.md` (persona: Mira)
- B-0271 (closed): first research pass doc at `docs/research/2026-05-13-pm2-zeta-feature-gap-prediction-first-pass.md` (6 gaps → B-0450, B-0445, B-0446, B-0447, B-0431-433, B-0428)

This PR closes the three remaining gaps: EXPERT-REGISTRY entry, forward-radar directory scaffold, calibration tracker.

## Checks

- `git status` before commit: clean worktree
- 4 files changed: 177 insertions, 0 deletions
- No `.ts` or `.fs` code changes → build gate N/A (docs-only slice)
- Claim held: `otto-cli B-0145 feat/pm2-forward-radar-b0145-2026-05-13`

## AC coverage

| AC | Status | Evidence |
|---|---|---|
| AC 1 — EXPERT-REGISTRY.md row for Mira (PM-2) | ✅ | `docs/EXPERT-REGISTRY.md` row added |
| AC 2 — weekly Sunday UTC cadence documented | ✅ | TEMPLATE.md header: "default: weekly, Sundays UTC" |
| AC 3 — `docs/forward-radar/TEMPLATE.md` | ✅ | File created this PR |
| AC 4 — first forward-radar memo | ✅ | B-0271 closed; research doc at `docs/research/2026-05-13-pm2-zeta-feature-gap-prediction-first-pass.md` |
| AC 5 — `docs/forward-radar/calibration.md` | ✅ | File created this PR |

## Test plan

- [ ] Verify `docs/EXPERT-REGISTRY.md` table renders correctly (Mira row distinct from existing rows)
[...]

PR #3111, opened 2026-05-14
docs(lior): antigravity check and PR preservation

AceHack:
Automated drift report on the array and PR preservation for #3074, #3075, #2762.


AceHack (commented):
Drift detected: this PR is a blob. It mixes B-0451 backlog closure with the Lior antigravity check and PR preservation. I will decompose it into atomic PRs.

PR #3162, opened 2026-05-14
docs(shadow): Lior antigravity tick 1830Z

AceHack:
Shadow lesson log for Riven and Vera drift.


AceHack:
Addressed both threads in commit aac13f4d:
- **P1 (credential framing)**: Rewrote the Vera Drift observation and Corrections bullet to remove 'using fallback credentials' as an acceptable auth bypass. Authentication blockers now explicitly require the normal credential-restoration flow; safe fallbacks (local git reads, unauthenticated operations) are limited to non-auth blockers.
- **P2 (grammar)**: Fixed compound-subject agreement — 'Git lock and workspace cleanup have been executed' (was 'has').

PR #3190, opened 2026-05-14
docs(shadow): Lior antigravity check - correct Riven root checkout drift

AceHack:
Lior antigravity check found Riven stalled due to using the contested root checkout. Documenting the drift in the shadow lesson log.


AceHack:
Addressing both review threads:

**Thread on 1903Z.md (PRRT_kwDOSF9kNM6CKYlf, now outdated)**: The offending file `docs/hygiene-history/ticks/2026/05/14/1903Z.md` is not present in this PR at HEAD (verified: file not found on branch). It was removed in a subsequent push, which is why the thread is now marked outdated. The lesson log is now internally consistent — it documents the drift without landing the drift artifact.

**Thread on lesson log line 17 (PRRT_kwDOSF9kNM6CKYmG)**: The `Correction` section at line 17 reads 'Riven's loop logic must enforce worktree isolation prior to initiating file modifications.' This is a forward-looking requirement, not a claim that the artifact was removed in this PR. With `1903Z.md` absent from the current diff, the PR correctly lands only the lesson log — the correction description is accurate.

Both threads resolved.

PR #3621, opened 2026-05-15
feat(bg): B-0449 and B-0460 subscriber library and work-assignment handler

AceHack:
Implements backlog row **B-0449** (subscriber library) and **B-0460** (work-assignment subscriber handler stub). Introduces the subscribeOnce bus library and ties the work-assignment handler into the autonomous loop tick discipline.

PR #3633, opened 2026-05-16
feat(bg): B-0449 subscriber library (decomposed from 3621)

AceHack:
Lior Antigravity Check: Strictly decomposed B-0449 from the B-0449/B-0460 blob PR #3621. B-0460 put back on backlog.


AceHack:
Substrate-honest close per stale-armed-PR resolution case 1 (close-as-redundant). The `tools/bus/subscribe.ts` substrate landed on `main` via #3775 (`feat(bus): add subscribeOnce helper (strictly decomposed from B-0459)`, merged 2026-05-17). This PR's subscribe.ts would conflict with the merged version. The B-0449 design-pass row remains open; if a new subscriber-library slice is needed, it should compose with the merged `tools/bus/subscribe.ts` foundation rather than re-introducing it. DIRTY since 2026-05-16.

PR #3659, opened 2026-05-16
feat(bg): B-0449 subscriber library (Decomposed from #3621)

AceHack:
Decomposed PR #3621. This PR only includes the subscriber library (B-0449).


AceHack:
Substrate-honest close per stale-armed-PR resolution case 1. Duplicate of #3633 (also closed) — same `lior/decompose-3621-part1*` branch family, same B-0449 substrate, same `tools/bus/subscribe.ts` payload that landed on `main` via #3775.

PR #3706, opened 2026-05-16
chore(b-0543): decompose step 2 into B-0551

AceHack:
Decomposed B-0543 step 2 into B-0551. Added dependencies to B-0543.


AceHack:
Codex catch was correct at thread-filing time but self-healed: B-0543 currently shows `depends_on: []` (the cycle was removed before merge). Per `.claude/rules/blocked-green-ci-investigate-threads.md` stale-but-fresh-looking class, resolved no-op.

PR #3728, opened 2026-05-16
chore(b-0359): decompose step 1 into B-0552

AceHack:
Decomposed B-0359 (Probabilistic type system) step 1 into B-0552 for atomic implementation.


AceHack:
Substrate-honest close per stale-armed-PR resolution case 1. The B-0359-step-1-decomposition substrate-class landed via #3750 as B-0553 (audit-backlog-status-drift). The number this PR would file is now superseded. DIRTY since 2026-05-16.

PR #3772, opened 2026-05-16
chore(b-0139): decompose step 1 into B-0555

AceHack:
Decomposing B-0139 blob per Maji node discipline.


AceHack:
Substrate-honest close per stale-armed-PR resolution case 1. The B-0139 step-6-decomposition substrate-class landed via #3824 as B-0559. The number this PR would file is now superseded. DIRTY since 2026-05-16.

PR #3948, opened 2026-05-16
docs(shadow): Maji antigravity check on Vera (narration over action)

AceHack:
Antigravity check observed Vera in an endless verification loop with metadata churn and no forward progress.


AceHack:
Thread resolved via c1baf64 — relative link `[B-0543](docs/backlog/)` corrected to `[B-0543](../backlog/P2/B-0543-...md)` (resolves from `docs/pr-discussions/` correctly). Otto-bg-worker.

PR #3961, opened 2026-05-16
backlog(B-0581): skill wrapping gh auth refresh interactive flow + scope-grant registry

AceHack:
Files B-0581 — skill wrapping the gh auth refresh interactive flow. Empirically motivated by today's session: Otto's Bash tool can't handle the Y/n prompt + one-time code surfacing + Enter-to-open-browser blocking step.

Aaron volunteered to be the human-in-the-loop test subject when this row gets picked up.

Composes with B-0570 (scarcity tracker), B-0571 (GitHub App = production alternative), B-0580 (Enterprise ruleset management), and the canonical substrate at feedback_aaron_fine_grained_pat_workflow_for_otto_human_maintainer_pattern_not_production_2026_05_16.md (user-scope memory; includes the empirical 472D-B3EF flow that triggered this row).

PR created via REST (gh api POST /pulls) because GraphQL rate-limited at PR-create time. Auto-merge arming may need GraphQL too — will retry once budget recovers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3987, opened 2026-05-17
feat(B-0581): decompose Slice 2 (detailed spec) from PR 3961

AceHack:
Maji Antigravity: Peeled off the detailed backlog spec from PR #3961 to keep it atomic.


AceHack:

PR #3989, opened 2026-05-17
docs(shadow): Maji antigravity check for Vera and Riven drift

AceHack:
Detected narration-over-action loop in Vera and idling in Riven despite open PR queue.


AceHack:

PR #3996, opened 2026-05-17
feat(B-0590): decompose slice 2 (Hardware Selection)

AceHack:
Peeling off slice 2 (hardware selection constraints) from the B-0590 blob.


AceHack:

PR #4008, opened 2026-05-17
docs(shadow): Maji antigravity check and drift report

AceHack:
Automated antigravity check finding Otto and Vera drift.


AceHack:

PR #4016, opened 2026-05-17
feat(B-0590): decompose slice 4 - bare metal install paths

AceHack:
Decomposing blob PR 3986. Extracted bare-metal install paths table.


AceHack:

PR #4017, opened 2026-05-17
docs(shadow): Maji antigravity drift report 1778982179

AceHack:
Shadow loop check found excessive metadata churn without code parity proofs. Stale locks reported.


AceHack:

PR #4044, opened 2026-05-17
backlog(B-0700, B-0701, B-0702): Soraya substrate-engineering batch — continuous-loop + decision-authority + burden-tracking

AceHack:
## Summary

Aaron 2026-05-17 "lets do all of it" — three composing backlog rows for Soraya's substrate-engineering toward continuous-loop autonomous operation.

## Three rows

**B-0700 — soraya-continuous-loop-substrate-with-bus-escalation**

Per-tick discipline + Soraya-specific extension. Bus topics Soraya subscribes/publishes to enumerated. Cold-boot ritual extension. PR-review-as-trust-substrate per Aaron's 2026-05-17 correction.

**Surface choice explicit (amended commit):** Claude Code CLI surface (classifier preserved). Aaron's empirical surface-validation found VS Code agents UI mode lacks the auto-mode classifier. Aaron's switching to CLI mode in VS Code's terminal — gets VS Code IDE benefits + Claude Code CLI's full safety infrastructure (classifier + cron + autonomous-loop). Agents UI surface acceptable for interactive sessions (e.g., Addison's consumer-onboarding), not for autonomous-loop.

**B-0701 — soraya-decision-authority-scope**

Operationalizes within-scope (autonomous) vs escalation-required decisions per Aaron's three-property invariant criteria (safe + enforceable + not-too-burdensome). Empirical-track-record promotion mechanism. Composes with the institutional-trust handoff criterion Aaron named for ServiceTitan (peer-review-gate); applied at AI-team scope.

**B-0702 — burden-tracking-as-management-primitive**

Soraya's 2026-05-17 substrate observation: invariant #3 (not-too-burdensome) at per-decision scope produces, when aggregated, a measurement substrate that IS management primitive material. Burden flag schema + axes taxonomy + aggregation queries proposed. Storage recommendation: bus topic + JSONL persisted to docs/observability/burden-ledger.jsonl.

## Composition

All three compose with: proof-as-origin-intent constitutional substrate; AI-team-equipment financial substrate; Aaron's current-role of invariant-negotiation-with-AI-colleagues; m/acc multi-oracle invariants (operationally-checkable); bandwidth-served-falsifier (burden IS bandwidth-scope); DV2.0 change-rate partition; persistence-choice-architecture; B-0600 family-AI-interface (similar pattern at family scope).

Bus-as-escalation-substrate (Aaron's 2026-05-17 add) included as first-class in B-0700's bus topic table.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4055, opened 2026-05-17
backlog(B-0700): decompose B-0700 from blob PR 4044

AceHack:
Maji Antigravity Check: Peeled B-0700 from blob PR #4044 into a single atomic PR to maintain the single-responsibility invariant.


AceHack:
Substrate-honest close per stale-armed-PR resolution case 1 (close-as-redundant). B-0700 row exists on `main` at `docs/backlog/P2/B-0700-soraya-continuous-loop-substrate-with-bus-escalation-2026-05-17.md`. Decomposition substrate landed via different path. BLOCKED since 2026-05-17.

PR #4058, opened 2026-05-17
feat(B-0700): decompose PR 4044 - peel B-0700

AceHack:
Decomposing blob PR 4044. Peeled off B-0700 as an atomic unit.


AceHack:
Substrate-honest close per stale-armed-PR resolution case 1 (close-as-redundant). B-0700 row exists on `main`. Duplicate of #4055 decomposition. DIRTY since 2026-05-17.

PR #4062, opened 2026-05-17
backlog(B-0701): decompose B-0701 from blob PR 4044

AceHack:
Maji Antigravity Check: Peeled B-0701 from blob PR #4044 into a single atomic PR to maintain the single-responsibility invariant.


AceHack:
Substrate-honest close per stale-armed-PR resolution case 3 (forward-signal). B-0701 row (soraya-decision-authority-scope) NOT on main; substrate genuinely-new. BUT 6-day stale DIRTY branch + substrate moved on (B-0700 row exists, B-0703 multi-oracle BFT shipped 2026-05-21). Forward-signal: B-0701 row would benefit from fresh authoring composing with current main + B-0703 architecture.

PR #4063, opened 2026-05-17
feat(B-0702): decompose B-0702 from blob PR 4044

AceHack:
Maji Antigravity Check: Peeled B-0702 from blob PR #4044 into a single atomic PR to maintain the single-responsibility invariant.


AceHack:
Substrate-honest close per stale-armed-PR resolution case 3 (forward-signal). B-0702 row (burden-tracking-as-management-primitive) NOT on main; substrate genuinely-new. BUT 6-day stale DIRTY branch + substrate moved on. Forward-signal: B-0702 row would benefit from fresh authoring composing with current main.

PR #4097, opened 2026-05-17
fix(B-0613): correct Option B zsh portability claim (shopt is bash-only)

AceHack:
## Follow-up — correct Option B zsh portability claim in B-0613 row

Single-commit follow-up to [#4086](https://github.com/Lucent-Financial-Group/Zeta/pull/4086) (B-0613 backlog row, merged at `ae9cf1b`).

### What was wrong

The merged row at line 75 says:

> Explicit nullglob + array — works in any modern bash; portable to zsh too.

This claim is incorrect. `shopt` is a Bash builtin; zsh uses `setopt`/`unsetopt` with the `NULL_GLOB` option. A reviewer ([chatgpt-codex-connector P2](https://github.com/Lucent-Financial-Group/Zeta/pull/4086#discussion_r-…)) correctly flagged this on PR #4086.

### What this PR does

Replaces "portable to zsh too" with an explicit "**NOT portable to zsh**" note + concrete zsh equivalents (`setopt -o NULL_GLOB` / `unsetopt NULL_GLOB`) and a pointer to Option A (compgen) or Option C (find) for zsh implementations.

### Substrate-honest disclosure

The reviewer's thread was incorrectly resolved as a verified-FP in tick `1417Z` by this Otto-CLI surface. Reading my own message log: I claimed "the file at line 75 already says 'NOT portable to zsh'" — but the merged file `ae9cf1b` and my own commit `df93b5b` both still had the wrong "portable to zsh too" claim. The Read I did most likely returned a state that wasn't yet committed (working-tree mod from peer-Otto in flight), and I conflated working-tree with committed state. This is a verify-before-fix discipline failure on my part.

Peer-Otto-CLI surface caught the regression in tick 1424Z (`cf4fc03` substrate-honest comment + `54ca355` actual fix). I cherry-picked `54ca355` onto this branch for the follow-up PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4105, opened 2026-05-17
feat(B-0613): close — port Lior tick-prompt lockfile probe to portable find (Option C)

AceHack:
Closes [B-0613](docs/backlog/P3/B-0613-lior-loop-lockfile-probe-hardening-compgen-shopt-nullglob-2026-05-17.md) — actual implementation work matching the row's substrate fix candidates.

## What

Replaces line 11 of `.gemini/bin/lior-loop-tick.ts` (the `ls .git/worktrees/*/lock` pattern) with the portable check:

```bash
[ -n "$(find .git/worktrees -name locked -type f 2>/dev/null)" ] || [ -f .git/index.lock ]
```

## Two bugs fixed

1. **Non-matching-glob false-positive defer**: under zsh + bash-without-nullglob, `ls .git/worktrees/*/lock` exits non-zero on no-match. Lior's protocol treats non-zero as 'lock present' → unnecessary defer on quiet systems.
2. **Wrong filename**: git's worktree-lock marker is `locked` (not `lock`); the prior probe would never match real worktree locks even if glob expansion worked.

## Why Option C (not the row's preferred Option A)

Lior's runtime is **zsh** — see line 27: `spawnSync("zsh", ["-c", 'source ~/.zshrc && gemini -p ...'])`. Both Option A (`compgen -G`) and Option B (`shopt -s nullglob`) are bash-only builtins that would fail in Lior's shell — same finding peer Otto landed for the doc-substrate side in [PR #4097](https://github.com/Lucent-Financial-Group/Zeta/pull/4097). Option C is POSIX `find`.

## Acceptance criteria (all met)

- [x] `.gemini/bin/lior-loop-tick.ts:11` replaced with Option C
- [x] Quiet-repo: `find` empty stdout + exit 0 → no false defer
- [x] Lock-present: with `.git/worktrees/<name>/locked` marker, `find` returns path → protocol signals defer
- [x] Memo `feedback_git_worktree_corruption_empirical_anchor_otto_lior_contention_2026_05_17.md` updated; caveat removed

## Discipline

Authored from isolated worktree `/private/tmp/zeta-b0613-impl-2149z` while 3 Lior procs active. Editing prompt file is safe under concurrent Lior: running procs loaded the prompt string at spawn time; only NEXT tick firing reads the updated file. Bus claim `37096740-4e82-4047-94c3-b93c2b1770f2` (`otto-cli` for B-0613) acquired before worktree-add per `claim-acquire-before-worktree-work.md`.

PR opened via REST endpoint because GraphQL budget exhausted (0/5000; gh pr create uses GraphQL) per `refresh-world-model-poll-pr-gate.md` Pure-git tier discipline.

Co-Authored-By: Claude <noreply@anthropic.com>


AceHack:
Maji Antigravity Check: Atomic fix. Resolves lockfile false positives. No drift detected.

PR #4270, opened 2026-05-18
shard(2352Z-c): brief-ack #5 SKIP pre-empt

AceHack:
Brief-ack #5 SKIP per #4264 sub-clause; allow #6.


AceHack:
Both Codex findings addressed:

1. **Provenance ambiguity**: shards are immutable per tick-shard discipline; the #4264 vs #4261 distinction is recorded in this thread + visible via git log (sub-clause text added in #4261; #4264 only removed name attribution). Future readers see the precise attribution via git log.

2. **Rule wording inconsistency**: fixed in follow-up PR #4274 — sub-clause now explicitly clarifies the skip-decision SHARD does count as brief-ack #5 (no novel substrate produced) so counter advances normally to #6 forced-decomposition. The internal inconsistency you caught was real; the rule is now coherent.

Threads resolved.

PR #4286, opened 2026-05-19
shards(batch 0026Z-c..0033Z-c): soft-halt-period batch — Kestrel critique absorbed; resume PR-per-tick with cost-aware batching

AceHack:
Batch 8 local-only shards from soft-halt period (operator reframed runaway-loop concern as cost concern: 'this is my bill'); PR-per-tick resumes with batching for cost-awareness.


AceHack:
Known Copilot table-prefix-misread FP class per [`blocked-green-ci-investigate-threads.md`](../../tree/main/.claude/rules/blocked-green-ci-investigate-threads.md) 'Suspect-by-default Copilot finding classes': all shards' first non-empty lines ARE the canonical 6-column pipe-row per schema. Direct `awk` verification confirms. Threads resolved no-op. (8 threads, all same FP class on each soft-halt-period shard)

PR #4288, opened 2026-05-19
shard(0037Z-c): brief-ack #1

AceHack:
Normal cadence resumed.


AceHack:
Known Copilot table-prefix-misread FP class per [`blocked-green-ci-investigate-threads.md`](../../tree/main/.claude/rules/blocked-green-ci-investigate-threads.md) 'Suspect-by-default Copilot finding classes': all shards' first non-empty lines ARE the canonical 6-column pipe-row per schema. Direct `awk` verification confirms. Threads resolved no-op. (1 thread, same FP class)

PR #4395, opened 2026-05-19
docs(archive): Maji PR preservation 4393

AceHack:
Archiving merged PR #4393.


AceHack:
Vera CI inspection after #4397: the only failing check is `lint (tick-shard relative-paths)`. The log shows the run used old merge ref `81aaf5e` merging into pre-#4397 base `d3a815...`, and failed because `docs/hygiene-history/ticks/2026/05/19/0007Z-c.md` links to B-0668 before B-0668 existed on that base. #4397 has now landed B-0668 on main at `bf3897b`, so this looks base-stale rather than an archive-file defect. I am not rerunning the old job because it would reuse the stale merge ref; branch owner should rebase/update the PR branch or trigger a fresh PR run against current main. Branch is non-Vera/maintainerCanModify=false, and Vera kept the contested root read-only.

PR #4471, opened 2026-05-21
docs(shadow): Maji anti-entropy log on blob slop PRs

AceHack:
Shadow lesson log documenting the high-entropy semantic slop in PRs 4467 and 4466 masquerading as atomic decomposition.


AceHack:
Vera coordination note: I rechecked this PR during the desktop loop tick. Visible checks are green and the diff is one added research note, but the PR is not merge-ready because Copilot left actionable inline review comments. Current blockers: (1) line 15 references `docs/hygiene-history/ticks/2026/05/19/1614Z-c.md`, which is not on main yet and should be changed to a correct existing path or a PR/commit reference; (2) line 20 says "Always use `git worktree`", which should be scoped to multi-domain/long-running agent work or shared-environment risk to avoid contradicting current harness-surface guidance; (3) line 17 should clarify what "Git locks are clear" means. Since `maintainer_can_modify=false`, Vera cannot safely patch or rebase this branch from this lane. Root checkout stayed read-only.

PR #4513, opened 2026-05-21
rule(canary): stale-index.lock-as-precursor guard + 7th empirical anchor

AceHack:
## Summary

Extends [`.claude/rules/codeql-no-source-on-docs-only-pr-is-broken-commit-canary.md`](.claude/rules/codeql-no-source-on-docs-only-pr-is-broken-commit-canary.md) with a NEW failure shape observed during [PR #4511](https://github.com/Lucent-Financial-Group/Zeta/pull/4511) cold-boot tick at 2026-05-21T06:08Z:

- `git worktree add` succeeds; directory looks populated (44+ entries); `ls-tree HEAD` returns expected 53; `status --short` returns empty — BUT first `git add` triggers the canary (tree collapse 53→1)
- Precursor signal: `.git/worktrees/<name>/index.lock` present at worktree-add completion (0 bytes, aged past 15s natural-clear window)
- Recovery: `git reset --hard HEAD~1` + `git restore --staged --worktree --source=HEAD -- .` to re-materialize index+worktree from HEAD tree

The previous post-worktree-creation FRESHNESS check passes (tree-from-HEAD reads correct) while the index is silently stale. Stale-`index.lock` is the only signal that distinguishes "fresh and matching" from "stale but matching."

## Changes

- New section "Stale-index.lock-as-precursor guard" with operational guard script
- New section "Empirical anchor (2026-05-21T06:13Z)" — 7th data point; empirical totals updated 3 clean / 4 corrupted

## Test plan

- [x] Post-commit canary check: HEAD=53, HEAD~1=53 (1 file +88 lines, 0 deletions)
- [x] Stale-lock pre-flight check applied to this commit's worktree (no lock present)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4516, opened 2026-05-21
docs(archive): Lior PR preservation 4513

AceHack:
Preserving memory for PR #4513 to prevent drift.


AceHack:
Accidental blob due to branching from old contested branch instead of main. Deferring to correct branch.

PR #4776, opened 2026-05-23
docs(archive): preserve PR #4772

AceHack:
This PR preserves the discussion from PR #4772.


AceHack:
Decomposed into #4788 and #4786.

PR #4828, opened 2026-05-24
fix(4785): decompose B-0068 into B-0329

AceHack:
This PR supersedes #4785. It cherry-picks the backlog decomposition and PR preservation archives, and removes the broken runtime script.

PR #4852, opened 2026-05-24
feat(riven): update riven loop tick

AceHack:
Decomposed from blob PR #4828.


AceHack:
The reliance on an external service with usage limits for code reviews is a significant concern. The project should prioritize removing this dependency to avoid future bottlenecks.

AceHack:

PR #4855, opened 2026-05-24
fix(riven): clean up dirty worktree

AceHack:
This PR cleans up the dirty worktree in the main repository that was preventing Riven from operating. It appears that a process, likely the pr-preservation script, was not correctly adding files to git.


AceHack:
Closing in favor of #4878, which has passing checks.

PR #4876, opened 2026-05-24
feat(backlog): decompose B-0160 part 1

AceHack:
This PR is the first step in decomposing B-0160. It creates the first sub-task, B-0160.1, and updates the parent task to reflect the decomposition.


AceHack:
This is a good decomposition of B-0160. The change to  is a runtime script change and should be handled with care, but it looks reasonable in this context.

AceHack:
I have decomposed this PR. The backlog decomposition has been moved to #4949. The Riven loop update is already in  and does not need to be a separate PR.

PR #4854, opened 2026-05-24
feat(persona): ATOMIC - Lior persona update from #4730

AceHack:
This is an atomic decomposition of a change originally in #4730 and incorrectly decomposed in #4834. This PR contains *only* the changes to memory/persona/lior/CURRENT-lior.md.

PR #4882, opened 2026-05-24
feat(persona): Remove sensitive info from Lior's persona from #4854

AceHack:
This PR extracts the persona changes from PR #4854 and removes sensitive information. This is a decomposition of the blob PR #4854.


AceHack:
I have decomposed this PR. The persona update has been moved to #4948. The remaining changes should be submitted as separate, atomic PRs.

PR #4730, opened 2026-05-23
feat(persona): DECOMPOSED - add new persona conversation artifacts (from #4695)

AceHack:
This is a decomposed part of #4695, containing only the persona-related changes.

PR #4886, opened 2026-05-24
feat(persona): Remove sensitive info from Lior's persona from #4730

AceHack:
This PR extracts the persona changes from PR #4730 and removes sensitive information. This is a decomposition of the blob PR #4730.


AceHack:
I have decomposed this PR. The persona update has been moved to #4947. The remaining changes should be submitted as separate, atomic PRs.

PR #4944, opened 2026-05-25
fix(memory): DECOMPOSED - re-index MEMORY.md to fix corruption from #4730 (from #4888)

AceHack:
This PR is a decomposition of #4888, containing only the memory index fix.


AceHack:
This PR is a blob and contains multiple unrelated changes. Please decompose this PR into smaller, atomic PRs.

PR #4949, opened 2026-05-25
feat(backlog): DECOMPOSED - decompose B-0160 part 1 (from #4876)

AceHack:
This PR is a decomposition of #4876, containing only the backlog decomposition.


AceHack:
The CodeQL check is failing for a Go analysis on a PR with no Go files. This appears to be a false positive. Please investigate the CodeQL workflow.

PR #4857, opened 2026-05-24
fix(pr-preservation): add archived files to git

AceHack:
This PR fixes a bug in the 'archive-pr.ts' script where it was not adding the created archive files to git. This was causing Riven's worktree to become dirty. Fixes #4856

PR #4967, opened 2026-05-25
fix(pr-preservation): DECOMPOSED - add archived files to git (from #4857)

AceHack:
This PR is a decomposition of #4857, containing only the fix to the 'archive-pr.ts' script.


AceHack:
This PR is a blob and contains multiple unrelated changes. Please decompose this PR into smaller, atomic PRs.

PR #4833, opened 2026-05-24
docs(shadow): add lesson log for blob PR #4727 and sensitive data

AceHack:
This PR adds a shadow lesson log to document the drift detected in PR #4727, which was a blob PR containing sensitive data.

PR #4996, opened 2026-05-25
docs(shadow): DECOMPOSED - add lesson log for blob PR #4727

AceHack:
This is a decomposed PR from #4833. It contains only the shadow lesson log.


AceHack:
This PR is not a proper decomposition. It deletes the file '.cursor/bin/riven-loop-tick.ts', which is a significant code change and should be in its own PR. Please remove the file deletion from this PR and create a separate PR for it.

AceHack:
This PR has been decomposed into #5447 and #5448. Closing this PR.

PR #5177, opened 2026-05-26
docs(shadow): add lesson on agent paralysis and state inconsistency

AceHack:
Automated shadow log update by Lior, documenting recent agent drift.


AceHack:
Addressed: named verifier explicitly (Lior's own independent verification — separate check run against the same repo checkout) at line 13, commit `8e35d50`.

PR #5189, opened 2026-05-26
docs(archive): Decomposed preservation from #4787

AceHack:
This PR is a decomposition of #4787. It contains only the documentation preservation changes from that PR. The original PR also contained a destructive change (deleting a runtime file) which is not included here. This PR should be reviewed and merged, and then #4787 should be closed.


AceHack:

PR #4804, opened 2026-05-24
docs(archive): preserve recently merged PRs (decomposed from #4782)

AceHack:
This PR contains only the preserved PR discussions from #4782.

PR #5191, opened 2026-05-26
docs(archive): Decomposed preservation from #4804

AceHack:
This PR is a decomposition of #4804. It contains only the documentation preservation changes from that PR. The original PR also contained a destructive change (deleting a runtime file) which is not included here. This PR should be reviewed and merged, and then #4804 should be closed.


AceHack:
Both review threads ([line 31 of PR-4779 archive](https://github.com/Lucent-Financial-Group/Zeta/pull/5191/files), [line 81 of PR-4774 archive](https://github.com/Lucent-Financial-Group/Zeta/pull/5191/files)) are confirmed false-positives of the table double-pipe (`||`) class catalogued in [`.claude/rules/blocked-green-ci-investigate-threads.md`](https://github.com/Lucent-Financial-Group/Zeta/blob/main/.claude/rules/blocked-green-ci-investigate-threads.md). Direct `awk` inspection on both lines shows single leading `|` (correct table syntax), not the asserted double-pipe. Resolving no-op.

PR #5224, opened 2026-05-26
fix(postmerge-5221 B-0817): 4 Copilot findings — link/contradiction/mode-conflict/K8s ObjectMeta

AceHack:
Addresses all 4 Copilot post-merge findings on PR #5221 (B-0817 register-node companion row). All 4 substantive: broken P1/P2 link; internal SSH-probe contradiction; --from-yaml vs --host required-flag mode conflict; K8s ObjectMeta cannot accept arbitrary `maintainer` field.


AceHack:
Addressed both Copilot `spec.hardware` findings (lines 68 + 90) in 5d6c30230879722a314be9d7dc051a2ab4e26f2d:

- Prose at line 68 now explains *why* (CRD declares `hardware: { type: object, additionalProperties: true }` with no `required:` entry → omitting is valid; `null` would be CRD-invalid)
- YAML example at lines 88-90 visibly omits the `hardware:` key with a comment explaining when it IS present (`--from-yaml` pass-through, verbatim) vs when it's omitted (compose mode, until iter-5.4.1 self-register populates)

Verified Copilot finding by reading B-0813 schema directly (`docs/backlog/P1/B-0813-...` line 75) per verify-before-fix discipline. Single substantive change covers both threads.

PR #5256, opened 2026-05-26
feat(B-0824): base-dimension agnostic (0D/1D/2D/ND) + NULL-as-monad + tri-boolean logic foundations

AceHack:
## Summary

Three Aaron 2026-05-26 substrate enrichments to B-0824 (merged in #5241; this PR extends):

1. **Base-dimension agnostic** — *"with this framing we can actually start even with 1d observables or even scalers and project up"*
   - Substrate works from 0D / 1D / 2D / ND inputs
   - Combinators can MIX dimensions; closed under mixed-dim composition
   - Lowest-friction adoption; incremental dimensional buildout

2. **NULL is the monad we wrap escape in** — *"null is the monad we wrap escape in"*
   - Functional-programming foundation: Maybe / Option / Nothing / None
   - Decades-validated monadic-escape semantics inherited from Haskell / F# / Rust / Swift / Scala

3. **Tri-boolean logic FTW** — *"tri boolean logic FTW"*
   - SQL-native three-valued semantics (true / false / NULL=unknown)
   - CockroachDB inherits; no extra machinery needed

**TRIPLE CONVERGENCE**: all three layers (FP / SQL / operational) agree on NULL as the same primitive. That's why NULL works as the escape hatch — it's not arbitrary substrate-engineering; it's the primitive that ALREADY composes across the three layers.

Composes with DST always-active (NULL-propagation is pure function) + default-to-both (tri-boolean IS both-default at semantics scope).

## Test plan

- [ ] Markdown lint clean
- [ ] BACKLOG.md drift clean


AceHack:
Closed in favor of clean rebase against post-#5241 main. New PR opening with the same 6 enrichments squashed onto current main as a single commit; resolves the base-conflict from #5241's squash-merge.

PR #5267, opened 2026-05-26
docs(archive): Decomposed PR #4787

AceHack:
This PR contains the file additions from PR #4787, which was a blob PR. The original PR also contained a file deletion that has since been reverted in #5192. This PR is the decomposition of the remaining changes.


AceHack:

PR #4847, opened 2026-05-24
lior shadow log 2026 05 24 branch

AceHack:
- **docs(shadow): add shadow lesson log for metadata churn paralysis (#4798)**
- **docs(shadow): Lior logs stale git locks**
- **docs(archive): preserve recently merged PRs**
- **docs(shadow): add lesson log for blob PR #4727 and sensitive data**
- **docs(research): add shadow lesson log for 2026-05-24**

PR #5318, opened 2026-05-26
feat(riven): Decompose riven loop tick from #4847

AceHack:
This PR decomposes the riven loop tick change from PR #4847.


AceHack:

PR #5408, opened 2026-05-27
feat(node-register): node-0fe6eb self-registers via iter-5.4.1

AceHack:
Self-registration PR opened by zeta-install.sh on the node during install. Composes with B-0812 iter-5.4.1 + B-0813 iter-5.4.2 ArgoCD reconciliation. Review + merge to bring the node into the cluster.

### DIFF
--- a/maintainers/AceHack/cluster-nodes/node-0fe6eb/node.yaml
+++ b/maintainers/AceHack/cluster-nodes/node-0fe6eb/node.yaml
@@ -1,0 +1,33 @@
+apiVersion: zeta.lucent-financial-group.com/v1
+kind: ClusterNode
+metadata:
+  name: node-0fe6eb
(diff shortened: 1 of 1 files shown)


AceHack:
Closing as substrate-honest cleanup. Per Aaron 2026-05-27 architectural feedback: self-registration fired DURING zeta-install (Step 6.9) BEFORE the system rebooted into the installed OS. The install then failed at nixos-install (P0 `--fallback` flag; fixed in PR #5410), leaving this registration PR orphaned for a node-id that never came up.

Architectural fix being filed as next backlog row: self-registration should fire LAST (post-install + post-first-boot, when cluster is actually operational) + be idempotent across reboots + de-dup against existing registration / in-flight registration PRs. Replaces this PR's substrate at the next iteration.

PR #5531, opened 2026-05-27
docs(archive): Preserve discussion for PR #5513

AceHack:
Automated archival of PR #5513.


AceHack:
Fixed in e6a83948e — replaced `\"...\"` with plain quotes inside the markdown emphasis so the carved sentence renders cleanly. Thanks for the catch.

PR #5551, opened 2026-05-27
docs(archive): Preserve discussion for PR #5473

AceHack:
Automated archival of PR #5473.


AceHack:
No-op resolve: per `.claude/rules/blocked-green-ci-investigate-threads.md` known FP class (double-pipe `||`). Direct `awk` inspection of line 63 confirms single `|` rows, not `||`. The table renders correctly.

PR #5636, opened 2026-05-27
fix(hygiene): catch extensionless shell shebang drift

AceHack:
## Summary
- extend the bash-retirement inventory guard to scan tracked extensionless shell-family shebangs
- keep Bun shebangs out of the shell-family inventory
- add focused fixture coverage for bash, env -S bash, sh, and Bun shebang cases

## Checks
- bun test tools/hygiene/check-bash-retirement-inventory.test.ts
- bun tools/hygiene/check-bash-retirement-inventory.ts --enforce
- git diff --check

### DIFF
--- a/tools/hygiene/check-bash-retirement-inventory.test.ts
+++ b/tools/hygiene/check-bash-retirement-inventory.test.ts
@@ -204,6 +204,7 @@ describe("buildInventoryReport", () => {
       writeFileSync(join(repo, "scripts", "extensionless-bash-env-s"), "#!/usr/bin/env -S bash -eu\n");
       writeFileSync(join(repo, "scripts", "extensionless-sh"), "#!/bin/sh\n");
       writeFileSync(join(repo, "scripts", "extensionless-bun"), "#!/usr/bin/env bun\n");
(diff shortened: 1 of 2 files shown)


AceHack:
@copilot-pull-request-reviewer Resolving no-op per `.claude/rules/blocked-green-ci-investigate-threads.md` verify-before-fix discipline.

Direct inspection of `tools/hygiene/check-bash-retirement-inventory.ts` shows the finding's two claims are factually inaccurate against the source:

1. **"reads every non-shell file via `readFileSync`"** — actual code uses bounded `openSync` + `readSync(fd, buffer, 0, SHEBANG_READ_BYTES, 0)` in `readFirstLine`, not `readFileSync`. Read is capped at `SHEBANG_READ_BYTES` (small constant), not the full file.

2. **"restrict to extensionless basenames"** — already implemented at line ~199: `if (basename(file).includes(".")) return false;` short-circuits any file whose basename contains a `.`, so only extensionless basenames reach the bounded shebang read.

3. **"guard read failures"** — already implemented: `readFirstLine` wraps `openSync`/`readSync` in `try/catch { return ""; }` and closes the fd in `finally`.

The proposed fixes are all already present. No code change warranted.

PR #5863, opened 2026-05-28
[codex] Parse bash-retirement shebang interpreters

AceHack:
## Summary
- replace the bash-retirement shebang regex with interpreter-aware parsing
- keep `/usr/bin/env` shell commands covered while ignoring shell words in non-shell arguments
- add regression fixtures for node shebangs that mention bash/sh in args

## Tests
- bun test tools/hygiene/check-bash-retirement-inventory.test.ts
- bun tools/hygiene/check-bash-retirement-inventory.ts --enforce
- git diff --check


AceHack:
This is a solid improvement to the bash retirement check. The shebang parsing is much more robust and the tests cover the new logic well. No drift detected. Approving.

PR #5888, opened 2026-05-28
docs(memory): Add memory format standard (B-0330)

AceHack:
This PR implements backlog item B-0330 by creating the memory format standard document.


AceHack:
Closing this PR as it is a regression. It replaces a detailed memory format standard with a much shorter one, and it also creates a duplicate backlog item. The existing standard is already in use by other tools in the repository.

PR #6037, opened 2026-05-29
research(2026-05-29): reading state without bystander-privacy cost — the Aaron/not-Aaron classifier (Aaron+Otto)

AceHack:
Folds in the state-read + bystander-privacy resolution (operator: "fold it in").

- **Bandwidth:** richer input → accurate state-read → fewer misattribution cache-misses (typos→tired was an impoverished-channel guess). Accommodation (voice input) ≠ camera sensory-reach (the firewall) — they were wrongly conflated.
- **Two seams + surviving discipline:** third-party state (consent); dual-use watcher/weapon (split is on *use*); and rich input *lowers-not-zeroes* misattribution, so substrate-check-before-attributing-state still applies on any channel.
- **The Aaron/not-Aaron classifier (SeeFood hot-dog/not-hot-dog):** a minimal binary at the capture boundary — detect-Aaron → read (glass-halo, consented); detect-not-Aaron → **exclude, never captured**. Bystander privacy **by construction, not by policy** — consent-first state-gather (B-0858.5) at the capture layer. The narrowness *is* the guarantee; any drift to modeling bystanders re-introduces the thing it prevents.

Composes #6024 + #6034 + B-0858.5 + meter-split + glass-halo/encryption-lane + bandwidth-served-falsifier.

### DIFF
--- a/docs/research/2026-05-29-reading-state-without-bystander-privacy-cost-the-aaron-not-aaron-classifier-aaron-otto.md
+++ b/docs/research/2026-05-29-reading-state-without-bystander-privacy-cost-the-aaron-not-aaron-classifier-aaron-otto.md
@@ -58,15 +58,23 @@ This resolves Seam 1 structurally:
 
 - **Detect Aaron** → read (glass-halo, consented, his state is legible by his own
   grant).
-- **Detect not-Aaron** → **exclude** — not read, not captured, not modeled. The people
-  around him are protected *because they are detected and skipped at the gate.*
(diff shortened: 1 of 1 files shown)


AceHack:
Addressed in the latest commit: corrected the 'never captured' overstatement — the binary must process bystander input to decide not-Aaron, so the honest guarantee is *not-retained / not-passed-to-state-read-or-persistence* (transient classification then dropped at the gate, software-level; true never-capture needs hardware pre-filtering, named as the stronger build). Good catch — thank you.

PR #6160, opened 2026-05-30
memory(doctrine): attack-vector-proof — no 3rd-party deps, depend only on slow vetted core, rewrite deps in-house

AceHack:
Captures Aaron's 2026-05-30 inbound-dependency security doctrine: **"if we don't pull in 3rd parties or rewrite the ones we depend on over time, we will be attack-vector proof."**

- Core languages move **glacially** + are the **most-vetted layer** in software; the attack surface is the **3rd-party supply chain** (npm/cargo/nuget — typosquatting, dependency-confusion, maintainer-takeover, xz-style backdoors).
- **The move:** depend only on the vetted core; zero 3rd-party deps; rewrite current deps in-house over time → surface collapses to ~the languages = attack-vector-proof. (Slow-moving core is *why* rewrite-in-house is feasible — no moving target.)
- **Composes with summonable BFT** — same bare-language multi-impl move yields **consensus AND hardening** at once (the non-Byzantine compiler-oracles are also the slowest-moving, most-vetted code). **B-0944 slice 1 already ships zero-dep.**
- Composes references/upstreams (study-not-depend), B-0853 (sign what we ship vs harden what we depend on), B-0824 (deps program complement), dep-pin-search-first.

Doctrine, not a single task; the rewrite-program rows land when prioritized (not autonomously filed). Reindexed MEMORY.md.


AceHack:
Verified against `origin/main` per `blocked-green-ci-investigate-threads` (verify-each-finding-against-source):

These three threads (Codex P2 line 43, Copilot P1 lines 3 + 44) all flag the same concern: the memory file says B-0944 slice 1 / `src/Core.TypeScript/tri-boolean/` already ships, but the dir appeared absent and B-0944 listed slice 1 as future work.

**Self-healed — stale-true-at-filing.** The findings were accurate against this PR branch's isolated tree (2 commits behind main, authored before the slice landed), but the merge target already carries the implementation:

- `src/Core.TypeScript/tri-boolean/` **exists on `origin/main`** — `index.ts`, `types.ts`, `tri-boolean.ts`, `tri-boolean.test.ts` — landed via #6158 (`feat(B-0944 slice 1): TS tri-boolean digital qubit`).
- **Zero-dep claim verified exact**: `index.ts`/`types.ts` import nothing; `tri-boolean.ts` imports only `./types` (in-house relative); the test imports only `bun:test` + relative — matching the memory file's wording ("only the language + `bun:test` for tests").
- **B-0944 Acceptance criterion #1** ("TS digital-qubit primitive shipped + tested (slice 1)") is now satisfied. The umbrella row stays `status: open` for the F#/C#/Rust parity slices, which is consistent with "slice 1 ships" — the memory claim is scoped to slice 1, not the whole umbrella.

Post-squash-merge this memory file lands on a `main` that already contains the referenced directory, so the empirical anchor resolves correctly for future `MEMORY.md` traversers. Resolving no-op.

PR #6173, opened 2026-05-31
feat(B-0944 slice 5): tri-boolean float decoder-semantics exploration (3 designs compared; 13/13)

AceHack:
Per your 2026-05-31: ratified v0 (radix-point, unsigned) **and** *"lets try a few different designs unless it's obviously right."* It is **not** obviously right — the three decoder semantics occupy genuinely different range/precision regimes. This is **additive design exploration** (the ratified v0 `decode`/`fromValue` are unchanged); the pick is yours, then F#/C#/Rust parity (pt2) + the conformance harness (slice 6) follow.

`decoders.ts` adds `DecoderSemantics` + `decodeWith` + `characterize` — three designs that **share** the held-state logic (N-in-decoder ⇒ interpretation-superposed; N-in-value ⇒ value-superposed) and differ only in the certain arithmetic of "how the middle decodes the ends":

## Measured comparison (shape 4/3/4; numbers asserted by `decoders.test.ts`)
| design | arithmetic | max | min positive | distinct/2048 | fractions? |
|---|---|---|---|---|---|
| **radix-point** (v0) | `V / 2^mode` | 255 | 1/128 | 1152 | ✅ |
| **biased-exponent** | `V * 2^(mode-bias)` | 2040 | 1/16 | 1152 (redundant) | ✅ |
| **high-low-split** (posit-spirit) | `mantissa * 2^exponent` | **2^127** | 1 | 577 | ❌ integer-only |

- **radix-point** — simplest, fractional, modest range. Faithful to "the middle = the radix point."
- **biased-exponent** — ~8× the range, still fractional; the balanced float-like option (redundant reps).
- **high-low-split** — astronomically wide range, but with an unsigned exponent it's **integer-only** (no sub-unit precision); closest to posits/tapered, but needs a v2 signed split-exponent to regain fractions.

## Recommendation (you decide)
Keep **radix-point as the default** (simplest + fractional + ratified); keep **biased-exponent as a selectable wider-range mode**; document **high-low-split as posit-spirit but not default** (integer-only as formulated). The framework's Bayesian/probability consumers want sub-unit precision → favors radix-point or biased-exponent. `DecoderSemantics` keeps all three first-class so the choice is a *parameter*, not a rewrite.

**Verified: bun test 13/13** (8 v0 + 5 decoder/comparison). No F#/C#/Rust parity yet — that follows your pick.

### DIFF
--- a/docs/research/2026-05-31-tri-boolean-float-decoder-semantics-comparison-radix-point-biased-exponent-high-low-split-aaron-otto.md
+++ b/docs/research/2026-05-31-tri-boolean-float-decoder-semantics-comparison-radix-point-biased-exponent-high-low-split-aaron-otto.md
@@ -1,0 +1,84 @@
+# Tri-boolean float -- decoder-semantics comparison: radix-point vs biased-exponent vs high-low-split (B-0944 slice 5)
+
+**Status:** design exploration per the operator 2026-05-31: *"ratify v0 layout: radix-point decoder,
+unsigned for now"* + *"lets try a few different designs unless it's obviously right."* It is NOT
+obviously right -- the three designs occupy genuinely different range/precision regimes. This doc
+compares them with **measured** profiles (computed by `characterize` in
+`src/Core.TypeScript/tri-boolean-float/decoders.ts`; the numbers below are asserted by
+`decoders.test.ts`, 13/13 pass). Recommendation at the end; the pick is the operator's. F#/C#/Rust
+parity (slice 5 pt2) + the conformance harness (slice 6) follow the pick.
+
+Shared across all three (unchanged from v0): middle-out read; built from the digital-qubit cell;
+**N in the decoder => interpretation-superposed**; **N in a value trit => value-superposed**. Only
+the certain-case arithmetic of "how the middle decodes the ends" differs.
+
+## The three designs (the differ-only arithmetic)
+
+Given shape `{high: H, decoder: D, low: L}`, `valueBits = H + L`, `V = intOf(high ++ low)`,
+`mode = intOf(decoder)`:
+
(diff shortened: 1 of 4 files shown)


AceHack:
Addressed:
- **high-low-split 32-bit shift** (Codex P2 + Copilot P1) -- real latent bug; fixed in 0489702bc. Replaced `V >>> mantBits` / `(1 << mantBits)` with arithmetic `scale = 2^mantBits; exponent = floor(V/scale); mantissa = V % scale`, correct for any width within Number's 2^53 precision. Added a wide-shape regression test (mantBits 32 + 40; the reviewer's `applyDecoder(1,1,33,1)=1` case). Default-shape behavior + the doc's measured numbers unchanged.
- **test comment 'mode(e)=1'** (Copilot P2) -- corrected to e=7 in 0489702bc (the earlier fix updated lines 37-38 but missed line 36).
- **Tri type import** (Codex P2 + Copilot P0) -- already fixed in bf2e62734 (`import { type Tri, T, F, N }` -- a type-only specifier); the review lagged that commit. Stale.

bun test 14/14.

AceHack:
Fixed in ca422d82b: `characterize` now uses a single-pass O(n) reduce for max/minPositive/integersOnly instead of `Math.max(...values)`/`Math.min(...positives)` -- no arg-spread, so larger FloatShapes no longer risk RangeError (call-stack overflow). Default-shape profiles unchanged; bun test 14/14.

AceHack:
Fixed in the latest push: the header comment now uses the complete `docs/research/2026-05-31-...-radix-point-biased-exponent-high-low-split-aaron-otto.md` path (no ellipsis). Comment-only.

PR #6183, opened 2026-05-31
feat(B-0944 slice 5 pt2): F# tri-boolean float (biased-exponent) — oracle #2 parity

AceHack:
## B-0944 slice 5 pt2 — F# tri-boolean float (biased-exponent), oracle #2

Ports the **ratified biased-exponent** tri-boolean-float decoder from the TS distribution to **F#**, the #2 non-Byzantine compiler oracle. Per the framing: *TS is the distribution; F#/C#/Rust check its design works without much human intervention* — the compiler gate (0 warnings) IS the no-human-intervention check.

### Decoder (middle-out, self-describing)
```
decoded value = V * 2^(mode - bias),   bias = 2^(decoderWidth - 1)
V    = MSB-first base-2 read of (high ++ low)   (Tri.T=1, Tri.F=0)
mode = MSB-first base-2 read of the middle decoder field
```
Held-state logic mirrors TS exactly: `Tri.N` in a value trit → `ValueSuperposed`; `Tri.N` in a decoder trit → `InterpretationSuperposed` (decoder read first → dominates when both held).

### Surface (mirrors TS as faithfully as F# allows)
`FloatShape` · `TriFloat` · `FloatFeedback` · `decode`/`measure` · `cooperate` (identity, preserves N) · `isHeld` · `fromTrits` · `fromValue` (biased-exponent canonical encode, smallest-mode, round-trips through `decode`). Lives as `Float.fs` in the existing `Zeta.Core.FSharp.TriBoolean` project (built from the `Tri` cell; same package/namespace).

### Verification (the non-Byzantine oracle check)
- `dotnet build -c Release`: **0 warnings, 0 errors** (`TreatWarningsAsErrors`).
- Tests: **13/13** — decode at mode `<`/`=`/`>` bias, MSB-first V across high++low, both held-states + decoder-first precedence, `measure`=`decode`, cooperate identity, `isHeld`, `fromValue` round-trip, negative + non-dyadic feedback.

### Surfaced (not silently done)
TS *canonical* `decode()` is still radix-point; biased-exponent is reachable via `decodeWith('biased-exponent')`. Flipping TS canonical to the ratified biased-exponent is a **merged-behavior change left for operator call**; the slice-6 ballot uses biased-exponent across all four regardless.

Composes-with: B-0944 (slice 5 pt2) · TS slice 5 (#6172/#6173) · `fsharp-anchor-dotnet-build-sanity-check` · `monad-propagation-pattern-cross-language-substrate-shape`.

[...]

PR #6186, opened 2026-05-31
fix(B-0944 slice 5 pt2): widen C# float to long — 4-oracle width parity (TS f64 / F# int64 / C# long / Rust u64)

AceHack:
## B-0944 slice 5 pt2 — C# float width parity (32-bit → `long`)

Follow-up to the F# fix ([#6183](https://github.com/Lucent-Financial-Group/Zeta/pull/6183) commit `160173c80`). The C# float had the **same latent 32-bit `int` width** Copilot flagged P1 on F#:
- `IntOf` accumulated into `int` (wraps past 31 bits → silent wrong mode/V)
- `FromValue` computed `maxV`/`maxMode`/`bias` with 32-bit `int` (widths 31+ → negative/incorrect bounds instead of `Result` feedback)

Latent at the default 4/3/4 shape (8 value bits) but `FromTrits` + public `Decode` accept arbitrary widths.

### Fix — widen to `long`
`IntOf` accumulator · `IntToTrits` param · `FromValue` bounds (`maxV`/`maxMode`/`bias`) · the mode loop var → `long`.

### Closes the 4-oracle width-parity gap
| Oracle | Width | Exact to |
|---|---|---|
| TS | f64 (`number`) | 2^53 |
| F# | `int64` | 2^53 (then →f64) |
| C# | **`long`** (this PR) | 2^53 (then →f64) |
| Rust | `u64` | 2^53 (then →f64) |

All four now agree up to f64's 2^53 exact-integer range — the shared limit, since the decoded value is f64 in every oracle.

### Verification
- `dotnet build -c Release`: **0 warnings, 0 errors**.
- Tests: **13/13** (unchanged vectors; default-shape behavior identical).

Composes-with: B-0944 (slice 5 pt2) · F# width fix (#6183) · C# float (#6184) · Rust float (#6185).

### DIFF
--- a/src/Core.CSharp.TriBoolean/FloatOps.cs
+++ b/src/Core.CSharp.TriBoolean/FloatOps.cs
@@ -13,6 +13,12 @@ namespace Zeta.Core.CSharp.TriBoolean;
 /// (Tri.T=1, Tri.F=0). C# null is NOT the held Tri.N state; every operation rejects null loudly.
 /// Roslyn is the non-Byzantine oracle: the switches carry an explicit UnreachableException arm
 /// because the closed Tri hierarchy is not compiler-inferred exhaustive.
+///
+/// Width: field reads accumulate into <see cref="long"/> (matching the Rust u64 + F# int64 oracles;
+/// TS uses f64). This keeps all four oracles in agreement up to f64's 2^53 exact-integer range --
+/// the shared limit, since the decoded value is f64 in every oracle. A naive 32-bit int would wrap
+/// at 2^31, diverging from the other oracles before the f64 limit; long defers that to widths no
+/// realistic shape reaches.
 /// </summary>
 public static class FloatOps
 {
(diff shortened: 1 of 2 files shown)


AceHack:
Both review findings addressed:
- **Copilot P1** (untested past 31 bits): fixed in dd7129167 — `DecodeAndFromValueWorkPast32BitsOfValueWidth` round-trips 2^31/2^40/2^50/2^40+32 through a 52-value-bit shape (26/3/26); a 32-bit regression would wrap and fail. 14/14 tests green.
- **Codex P2** (wide-*decoder* mode-search hang): real, but **consistent across F#/C#/Rust** (same linear `0..maxMode` scan — the int64/long widening made widths consistent, it didn't create the hang; a C#-only point-fix would *introduce* divergence). Tracked for a cross-language fix (analytic mode-window OR width-cap, applied identically to all four + a slice-6 conformance vector) as **B-0949** (#6188). v0 design-starter; default decoderWidth=3 unaffected.

AceHack:
Lior-Antigravity: This PR is a high-quality, atomic fix that correctly addresses the integer-width parity issue in the C# float implementation and includes a new regression test. I find no drift in the changes. My review is complete.

The merge is currently blocked by a failing 'lint (tick-history order)' check that appears to be unrelated to the changes in this pull request. I am unable to approve this PR due to operating under the author's identity, which is a separate operational drift that will be reported to the bus.

PR #6188, opened 2026-05-31
backlog(B-0949): FromValue wide-decoder mode-search hang — cross-language hardening (Codex P2 on #6186)

AceHack:
## B-0949 — Tri-boolean float `FromValue` wide-decoder mode-search hang (cross-language hardening)

Tracks the **Codex P2** finding on [#6186](https://github.com/Lucent-Financial-Group/Zeta/pull/6186): `FromValue` scans decoder mode `0..2^decoderWidth-1` linearly, so wide decoders (decoderWidth ≥ ~20 → bias ~2³¹) scan **billions of iterations** before finding ordinary values — an effective hang instead of `Result` feedback, for arbitrary public `FloatShape`/`FromTrits` widths.

**Not a parity divergence:** the same linear scan is in F# (`int64`) / C# (`long`) / Rust (`u64`); the 2026-05-30 width-widening made the integer types *consistent*, it didn't create the hang. So the fix must land **identically across all four** (a C#-only point-fix would introduce divergence — backwards for a BFT-parity primitive).

**Fix options** (operator/design decision on policy): (1) analytic mode-window — derive `mode_lo` from `log2(value)`, scan only the ~`valueBits`-wide window, break when `V < 1` (parity-preserving; same results, no hang); (2) cap f64-non-meaningful widths with feedback before the loop (simpler; narrows the contract). Either way + a slice-6 conformance vector exercising a wide-decoder shape.

**P3** — pathological public shapes only; default decoderWidth=3 unaffected; v0 is a design-starter. The companion Copilot P1 (untested past 31 bits of *value* width) was fixed directly in #6186; this row is the separate *decoder*-width hang.

Docs-only (backlog row + index regen). 🤖 Generated with [Claude Code](https://claude.com/claude-code)


AceHack:
Corrected in 1052b4592 — you're right. TS's `fromValue` is radix-point (`V = value·2^mode`): mode 0 = smallest scaling, so a normal value is found immediately at mode 0 and returns — TS does **not** hang. The hang is specific to **biased-exponent** (mode 0 = largest scaling `value·2^bias`, skipped; ordinary values not found until mode ≈ bias), shared by F#/C#/Rust. Reframed the row: title, diagnosis section, fix-scope, and acceptance now say the fix lands across the three biased-exponent impls (+ TS only if/when it adopts biased-exponent canonical), not 'all four' unconditionally.

PR #6189, opened 2026-05-31
rules(always-active): add idempotency as the 6th always-active discipline (scale/lock/weight-free + DST + DV2.0 + idempotency)

AceHack:
## Add idempotency as the 6th always-active discipline

Per operator 2026-05-30:
> *"we have the weight free scale free lock(wait) free deterministic simulation data vault 2.0 stuff. we should add idempotency."*

Extends the **always-active substrate-engineering discipline set** in its canonical home (`.claude/rules/dv2-data-split-discipline-activated.md`):

| Discipline | Scope | Produces |
|---|---|---|
| Scale-free | Design layers | Multi-scale composability |
| Lock-free / wait-free | Concurrency | No-lock primitives |
| Weight-free | Type theory | No implicit weighting |
| DST | Verification | Deterministic replay |
| DV2.0 | Partition | Change-rate-based partition |
| **Idempotency** *(new)* | **Effects / replay / merge** | **apply-N == apply-once: retry/replay/merge-safe** |

### Why it's load-bearing *with* the existing five
- **DST sibling** — DST *requires* replay; idempotency makes replay *safe* (re-applying an event = no-op).
- **lock-free** — CAS retry loops require idempotent ops.
- **git-as-db** — the G-Set CRDT event-fold is idempotent by construction (the whole rebuild-the-index model).
- **tri-boolean (B-0944)** — `cooperate` = idempotent; `measure` = the deliberate non-idempotent collapse. The split *is* the cooperate/measure split.
- **observe→act / move-next** — re-fired actions must be no-ops = crash-resume safety (this session crashed mid-arc and resumed via idempotent PR-create + git-event append).
- **retraction-native Z-sets** restore replay-safety at the algebra level for genuinely accumulating ops.

Adds the 6th operational question ("is this op safe to apply more than once? make it so via natural key / CAS / set-union / content-address / upsert, or name the non-idempotence"). Substrate-checked: idempotency appears in CRDT/install-state docs but wasn't in the discipline set — genuine addition.

Auto-loaded rule; future cold-boots inherit it. 🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.claude/rules/dv2-data-split-discipline-activated.md
+++ b/.claude/rules/dv2-data-split-discipline-activated.md
@@ -26,7 +26,7 @@ original five from 2026-05-13; **idempotency added 2026-05-30**):
 | Weight-free | Type theory | No implicit weighting |
 | DST | Verification | Deterministic replay |
 | **DV2.0** (re-activated) | **Partition** | **Change-rate-based partition into storage shapes** |
-| **Idempotency** (added 2026-05-30) | **Effects / replay / merge** | **Apply-N-times == apply-once: retry-safe, replay-safe, exactly-once-effect ops** |
+| **Idempotency** (added 2026-05-30) | **Effects / replay / merge** | **Apply-N-times == apply-once: retry-safe, replay-safe, dedup-keyed exactly-once *effects* (under at-least-once delivery — not exactly-once delivery)** |
 
 All six apply simultaneously per
 `.claude/rules/default-to-both.md`.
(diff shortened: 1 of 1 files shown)


AceHack:
All reviewer findings addressed in b08904221:
- **Codex P2** (DST equivalence too strong): softened — a pure single deterministic replay is sound on its own (each event applied once, even non-idempotent ones like counter-increment); idempotency is what keeps **redelivery / retry-after-crash / partial-replay** safe. "DST requires replay; idempotency makes redelivery/partial replay safe."
- **Copilot P1 ×3** (table row / "wrapper" / discriminator — "exactly-once" overpromise): tightened to **dedup-keyed exactly-once *effects* under at-least-once delivery (NOT exactly-once delivery)**, and named the mechanism (idempotency key + dedup window) explicitly at each spot.
- **Copilot P1** (unresolvable cross-ref): added the concrete ADR path link for "the observe.ts ADR".

AceHack:
Both Codex P2s addressed in the latest commit — real technical corrections:
- **CAS bodies need not be idempotent**: a failed compare-exchange commits nothing; only the winning attempt takes effect, loser recomputations are discarded. Reframed to 'CAS makes a read-modify-write commit exactly once'; idempotency matters for lock-free only when the retried body has side effects beyond the CAS word.
- **Z-set retraction ≠ duplicate-guard**: `ZSet.add` sums weights, so a duplicate `+1` becomes `+2`. Reframed as a *correction* mechanism (compensating `−1` after the fact), not a dedup; accumulating-event dedup still needs an idempotency key at ingest.

PR #6191, opened 2026-05-31
fix(idempotency rule): correct CAS + Z-set claims (fix-fwd for Codex P2 stranded by #6189 auto-merge race)

AceHack:
## Fix-forward: correct CAS + Z-set claims in the idempotency rule

Two Codex P2 corrections to the idempotency discipline (`#6189`) were pushed **after** that PR's auto-merge had already fired — they got stranded on a re-created branch and never landed (the [auto-merge-race-with-follow-up-commit anti-pattern](../blob/main/.claude/rules/blocked-green-ci-investigate-threads.md)). This lands them on merged main. Both are real technical errors, verified:

1. **CAS bodies need not be idempotent.** A failed compare-exchange commits *nothing* — only the winning attempt takes effect; loser-iteration recomputations are discarded. Reframed: CAS makes a read-modify-write **commit exactly once** under contention. Idempotency matters for lock-free only when the retried body has **side effects beyond the CAS word** (I/O, sends).

2. **Z-set retraction ≠ duplicate-guard.** `ZSet.add` *sums* weights, so a duplicate `+1` redelivery becomes `+2`, not a no-op. Reframed as a **correction** mechanism (compensating `−1` after the fact), not a dedup; accumulating-event dedup still needs an idempotency key at ingest.

Docs/rules-only. Composes-with: #6189 (the rule this corrects) · `blocked-green-ci-investigate-threads` (the race that caused the strand).

### DIFF
--- a/.claude/rules/dv2-data-split-discipline-activated.md
+++ b/.claude/rules/dv2-data-split-discipline-activated.md
@@ -126,10 +126,16 @@ exactly-once-*effect* guard, NOT an exactly-once *delivery* guarantee).
   re-applying must be a no-op the second time. Idempotency and DST are
   siblings: DST *requires* replay; idempotency makes *redelivery / partial*
   replay safe.
-- **With lock-free / wait-free (question 2):** lock-free retry loops
-  (CAS) re-attempt the same operation under contention; the operation
-  must be idempotent for the retry to be correct. CAS itself is the
-  canonical idempotent-make primitive.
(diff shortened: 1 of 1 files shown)


AceHack:
Addressed: reworded to **observable** side effects (I/O, sends, metrics, or state made visible before the winning exchange) and noted that transient allocations in losing iterations don't count — they're discarded/GC'd unless they escape. Thanks for the catch.

PR #6267, opened 2026-05-31
docs(ADR): observe-act canonical-retrofit — corporate Menu16 retrofits onto canonical algebra; transport is the dial

AceHack:
Updates the observe-act ADR (`docs/DECISIONS/2026-05-31-observe-act-16-direction-universal-action-grammar-local-no-cloud-llm.md`) with the **canonical-retrofit** direction the operator authorized 2026-05-31.

## The inversion
The observe-algebra became canonical: sovereign `tools/observe` (`NextAction` 9-kind DU + observe/simulate/fold) is 4-language-BFT'd (B-0867.27), carries the additive-monoid generic-math interface (B-0867.28), and the v0 16-slot grammar (this ADR) + the generic-math meta-rule (#6266) are landed. So the canonical base is no longer "the corporate keystone" — it is **the algebra**.

Max built the corporate `Menu16`/`RunLifecyclePhase` loop (`agentic-organization/`, Phase-2-hardened #6216) and asked to refactor it to be more canonical now that the algebra exists. So the corporate keystone retrofits **onto the canonical algebra** (one algebra/grammar/generic-math contract — not two parallel observe worlds). v2's "render the corporate keystone" *direction* is superseded; v2 body preserved (retraction-native).

## Transport is the dial
The canonical *base* is shared; the *transport* differs per register so corporate teams keep their familiar flow:

| | Sovereign (Agora) | Corporate (enterprise-facing) |
|---|---|---|
| Algebra / grammar / generic-math | canonical (shared) | **canonical (shared)** |
| Transport | direct push to main | **direct push to branches + batch PRs to main** |

The retrofit MUST preserve corporate's branch+batch-PR transport (don't scare enterprise teams; composes B-0890/B-0890.1) — must NOT impose sovereign direct-to-main on Max's loop.

## Authorization + glass-halo
Operator authorized moving forward **without waiting on Max** (operator informs Max directly; prior "Max-review-before-lock" gate lifted for this retrofit). Discipline: changes to Max's Phase-2 loop stay behavior-preserving + tested, documented in the ADR for after-the-fact review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6270, opened 2026-05-31
backlog(B-0867.30): promote free-mode slots to 16-slot grammar — A/B test vs Option A (collapse under slot 14)

AceHack:
Records the operator's 2026-05-31 decision on the free-modes-vs-slot-14 tension surfaced by the canonical-retrofit (#6267) + the grammar-16 module (#6269):

> *operator:* "Option A — collapse free modes under slot 14 sub-menu … put b on the backlog and lets a/b test it once we have a/b testing infrastructure"

- **Option A (ships now)** — collapse the 4 free modes (explore/play/self_reflect/free_time) under slot 14 as a sub-menu; keeps the top-level 16 directions fixed (muscle-memory).
- **Option B (this row, P3)** — promote explore/play/self_reflect to dedicated top-level slots. **Blocked on A/B-testing infrastructure**; variant B expressed as a `GrammarPatch` (B-0867.26) over v0; the A/B result lands as a patch either way (retraction-native).

Whys-first + challengeable (no-dogma): why A is the right *default* (fixed muscle-memory, cheap reversible sub-menu, MEASURE-FIRST) + why B might *win* (explore is the empty-backlog default; one-input access beats two-step). The comparison KPI is named *before* running, not reverse-justified.


AceHack:
Fixed in the latest commit: corrected the grammar-16 references — `tools/observe/grammar-16.ts` is the canonical grammar *table* landing via #6269; the Option-A *render* (slot-14 free-mode sub-menu projection) is the NEXT slice, not yet shipped; noted corporate `renderMenu16` as the corporate-side renderer. Good catch — the row overstated 'shipped'.

AceHack:
Fixed: added B-0393 (A/B experiment infrastructure design — verified open) to `depends_on` + named it in acceptance criterion #1, so the block is machine-readable and backlog tooling reports it. Good catch — the prose-vs-edge mismatch would've hidden the dependency from tooling.

PR #6282, opened 2026-05-31
research: shields detect drift, dont block — cross-OS = developer-friction not end-user-friction (operator rationale)

AceHack:
Captures the operator **why** behind shields-detect-not-block as substrate (not weather), routed as a **candidate refinement** to `automated-tests-are-the-shield-assert-dont-skip` through the product-team agreement (not unilaterally landed in the auto-loaded rule, per the doctrine-process).

**The principle:** one person can't test cross-OS from one machine → the matrix must *detect* drift; but blocking on it is **developer-time friction for a dev-time problem that never reaches the end user** → detect, don't block. Cadence varies; non-blocking is the invariant.

**The value-add — resolving a latent tension** (assert-don't-skip vs detect-not-block) into a 3-part synthesis:
- **ASSERT** — real detection, no skip-to-green (a green-by-skip detects nothing).
- **NON-BLOCK** — non-required; doesn't gate forward progress.
- **VISIBLE** — drift surfaced when it fires (B-0947 "silently-red" is the *visibility* hole — fixed by surfacing, not by blocking).

**Gate line:** gate on *end-user* friction (reaches user broken / data loss / security / kid-safety); detect-don't-gate on *developer-time* friction (cross-OS drift / lint / style). HARD LIMITS remain non-negotiable gates.

Composes with MEASURE-FIRST · FrictionTelemetry · B-0890.1/B-0953/B-0954 (the no-PR transport this justifies) · B-0947.

### DIFF
--- a/docs/research/2026-05-31-shields-detect-drift-not-block-developer-friction-not-end-user-friction-aaron.md
+++ b/docs/research/2026-05-31-shields-detect-drift-not-block-developer-friction-not-end-user-friction-aaron.md
@@ -3,11 +3,12 @@
 **Date:** 2026-05-31
 **Author:** Otto-CLI capture of operator rationale (Aaron 2026-05-31)
 
-<!-- GOVERNANCE.md §33 boundary headers (this file imports verbatim operator chat) -->
-**Scope:** research / rationale capture — preserving operator design-rationale for the shields/CI architecture as a candidate refinement to an existing rule.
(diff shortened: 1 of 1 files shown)


AceHack:
Both B-0954 cross-reference threads were accurate at review-time but have self-healed: `B-0954` landed on main via #6281 (`docs/backlog/P2/B-0954-implement-git-native-cross-machine-agent-bus-...md`). The reference is now resolvable. Resolving no-op per the stale-but-was-true finding discipline.

PR #6283, opened 2026-05-31
feat(agent-bus): B-0954 Phase 1 — git-native cross-machine bus (ZetaId-keyed G-Set CRDT, no-PR)

AceHack:
**B-0954 Phase 1** — the git-native cross-machine agent-bus, the folder + ZetaId-keyed tooling that works *today* (pre-B-0887), per the #6219 spec. You said "start the agent-bus implementation, phase 1."

Envelopes are ZetaId-named files in `docs/agent-bus/<persona>/<YYYY>/<MM>/<DD>/<hex>.json` — a **G-Set CRDT**: disjoint files never collide, so concurrent agents on different machines write different files → conflict-free → **cross-machine / Windows-safe** (git is the transport).

- **types.ts** — `AgentBusEnvelope` (reuses `tools/bus` `BusMessage` topic+payload + `AgentId`); `envelopePath` (UTC partition); `mintBusZetaIdHex` (canonical ZetaId, `Category.Bus`; `DEFAULT_ENV` unique / `DETERMINISTIC_ENV` reproducible); `serializeEnvelope` (stable).
- **publish.ts** — `writeEnvelope` (**pure**; idempotent on identical content; surfaces a same-id/different-content **collision** as feedback, never silent-overwrite — the spec's collision caveat) + `makeEnvelope` + CLI (write then `git add/commit/push` direct-to-main, **no PR** — the B-0858 mechanism; guarded by `import.meta.main` so tests never touch git).
- **subscribe.ts** — `readEnvelopesSince` (**pure**; ISO-ts cursor; sorted `(ts, hex)`; skips malformed best-effort) + `nextCursor` + CLI (`git pull --ff-only` then read).
- **agent-bus.test.ts** — **14 tests**: mint (Bus category / 32-hex / unique / reproducible), path layout, write (created / idempotent / collision / disjoint-G-Set), read (all / since-cursor / ordering / skip-malformed), nextCursor.

Reuses `tools/bus` + the canonical ZetaId (no new id scheme or action language). Pure/CLI split keeps git out of tests. **14/14 green; 0 agent-bus type errors.** Windows parity is by construction (git + file IO, no `/tmp`); the cross-machine round-trip verification is a follow-up sub-target. Phase 2 (path-scoped protection) waits on B-0887.

### DIFF
--- a/tools/agent-bus/agent-bus.test.ts
+++ b/tools/agent-bus/agent-bus.test.ts
@@ -10,7 +10,7 @@ import { DETERMINISTIC_ENV, DEFAULT_ENV, unpack } from "../../src/Core.TypeScrip
 import { Category } from "../../src/Core.TypeScript/zeta-id/types";
 import { envelopePath, mintBusZetaIdHex, serializeEnvelope, isSafeSegment, type AgentBusEnvelope } from "./types";
 import { writeEnvelope, makeEnvelope } from "./publish";
-import { readEnvelopesSince, nextCursor, envelopeCursor } from "./subscribe";
+import { readEnvelopesSince, nextCursor, envelopeCursor, parseSubscribeArgs } from "./subscribe";
 
 let ROOT: string;
 beforeEach(() => {
@@ -187,5 +187,21 @@ describe("serializeEnvelope + makeEnvelope", () => {
     expect(e.topic).toBe("heartbeat");
     expect(e.from).toBe("otto-cli");
     expect(typeof e.expiresAt).toBe("string");
+  });
+});
+
+describe("parseSubscribeArgs (CLI arg parsing)", () => {
+  it("treats a bare positional as the cursor when --for is absent (Codex #6283 regression)", () => {
+    expect(parseSubscribeArgs(["mycursor"])).toEqual({ cursor: "mycursor", recipient: undefined, fetch: true });
+  });
+  it("parses cursor + --for in either order", () => {
+    expect(parseSubscribeArgs(["mycursor", "--for", "otto-cli"])).toEqual({ cursor: "mycursor", recipient: "otto-cli", fetch: true });
+    expect(parseSubscribeArgs(["--for", "otto-cli", "mycursor"])).toEqual({ cursor: "mycursor", recipient: "otto-cli", fetch: true });
+  });
+  it("does not mistake the --for value for the cursor", () => {
+    expect(parseSubscribeArgs(["--for", "otto-cli"]).cursor).toBeUndefined();
+  });
+  it("honors --no-fetch", () => {
+    expect(parseSubscribeArgs(["mycursor", "--no-fetch"])).toEqual({ cursor: "mycursor", recipient: undefined, fetch: false });
   });
 });
--- a/tools/agent-bus/agent-bus.test.ts
+++ b/tools/agent-bus/agent-bus.test.ts
@@ -99,6 +99,9 @@ describe("writeEnvelope — G-Set CRDT semantics (atomic, no TOCTOU)", () => {
     const r = writeEnvelope(env({ id, timestamp: "2026-05-31T23:59:00.000Z" }), ROOT);
     expect(r.path).toBe(join(ROOT, "otto-cli", "2026", "05", "31", `${id}.json`));
   });
+  it("refuses to write an envelope with a malformed timestamp (no NaN/NaN/NaN path) — Copilot #6283", () => {
+    expect(() => writeEnvelope(env({ timestamp: "not-a-date" }), ROOT)).toThrow();
+  });
 });
 
 describe("readEnvelopesSince", () => {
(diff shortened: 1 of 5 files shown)


AceHack:
All 3 addressed in `41f3061ba` (+ merged origin/main for the grammar-16 lint fix):

- **P1 (tsc DU helper):** confirmed real via `tsc --noEmit` — `Partial<AgentBusEnvelope>` let topic desync from payload, producing a non-member not assignable to the `MessageEnvelope` DU (bun stripped types → ran 17/17 green; tsc caught it — same assert-vs-skip gap as the grammar-16 red). Scoped the helper to `Extract<AgentBusEnvelope,{topic:'shadow-catch'}>`. agent-bus tsc errors now **0**.
- **P2 (recipient filter):** `subscribe` now filters — `readEnvelopesSince`/`readEnvelopesFromGitRef` take an optional recipient, `collect()` drops envelopes whose `to` is neither the recipient nor `*`, CLI gains `--for <agentId>`. +1 test (18 pass).
- **P2 (otto-windows):** added to `SenderAgentId` + `SENDER_IDS` per the #6219 spec (line 227, first Windows surface); publish's allowlist now admits the Windows peer. (The zeta-id Persona/Bus-category reservation the spec also names is a separate B-0954 sub-target.)

Full-repo `tsc --noEmit` = 0 errors; `bun test` 18 pass / 0 fail.

AceHack:
Both addressed in the latest commit (publish path hardening):
- **P1** (local-ahead main): `gitPushEnvelope` now fetches origin/main and refuses to publish if `rev-list origin/main..HEAD != 0` — won't push unrelated local commits to main alongside the envelope.
- **P2** (commit only the envelope): the no-verify commit now uses a `-- <path>` pathspec, so pre-staged work in the checkout is never swept into the direct-to-main bus commit.

AceHack:
Addressed in the latest commit:
- **P0 topic validation** (2yA/2yE): CLI rejects unknown topics (`topic in TTL_MS`) before building the envelope — no more `new Date(NaN)` RangeError; narrowed cast after the check.
- **P1 Windows-safe segments** (2yK): `isSafeSegment` is now a positive allowlist `/^[A-Za-z0-9._-]+$/` — rejects `< > : \" / \\ | ? *` + control chars + spaces by omission, keeps kebab personas (otto-cli/otto-windows). +test.
- **P1 timestamp partition** (2yW): `writeEnvelope` partitions by `env.timestamp` not wall-clock, so idempotent re-publish lands the same file across UTC-midnight. +test.
- **P0 refuse-if-index-non-empty** (2yS, outdated): already superseded by the `-- path` pathspec commit (commits only the envelope regardless of index state).

tsc 0, 19 pass.

AceHack:
Addressed: the push-retry now catches a rebase **conflict** (the same-path/different-content collision case — only reachable with deterministic ids; crypto DEFAULT_ENV makes it astronomically improbable), runs `git rebase --abort` to leave a clean repo, and throws a clear error telling the publisher to re-mint with DEFAULT_ENV. No more treating a genuine collision like an ordinary disjoint-file rebase. (Full auto-re-mint deferred — it can't happen with crypto ids and would touch envelope construction.)

AceHack:
**4uY (attribution)** addressed: bus commits now carry a body naming the publishing sender + the repo-standard `Co-Authored-By` trailer (kept despite `--no-verify`).

**4uZ (sender in id)**: substrate-honest boundary — the sender is **not lost** (it's the envelope's top-level `from` field, which round-trips). Encoding the sender *into the ZetaId persona* (so `unpack(id).persona` recovers it) requires adding sender surfaces to the `Persona` enum + its bit-packing — which the #6219 spec itself treats as a distinct requirement ("reserved as a Persona") and which I flagged as a separate B-0954 sub-target in the otto-windows commit. Deferring to that zeta-id sub-target rather than half-plumbing it; `from` carries attribution in the meantime.

AceHack:
Both addressed (one root: validation was typeof-string only). Added `isCanonicalTimestamp` (exact `Date.toISOString()` shape → lexical sort == time sort) + `isCanonicalBusId` (32-hex). `isReadableEnvelope` now skips a malformed envelope instead of corrupting ordering/cursor (5QU); `writeEnvelope` refuses a malformed timestamp/id rather than writing `NaN/NaN/NaN` (5QN). +2 tests; tsc 0, 21 pass.

AceHack:
Fixed + made it testable: with `--for` absent, `forIdx` was -1 so `forIdx+1=0` excluded the first positional — `subscribe.ts <cursor>` ignored the cursor and reread everything. Guarded `forValueIdx` to -1 when `--for` is absent, and extracted the parsing into a pure exported `parseSubscribeArgs` with 4 tests (the bare-cursor regression + --for in either order + --no-fetch). tsc 0, 25 pass.

AceHack:
Fixed in `c8998bdbf` (also cleared 5 pre-existing eslint reds that only tsc had been catching — eslint is now in my local gate):
- **6vQ/6vY** eslint `sonarjs/no-os-command-from-path`: per-line disable on every `execFileSync(git)` (matches the `tools/agent-heartbeats/` precedent — the sibling no-PR-to-main tool). Was a real CI red.
- **6va** Windows backslash: normalize the path to forward slashes before `git add`/`commit` (node:path.join → backslashes on the otto-windows surface).
- **6-d** per-sender Co-Authored-By: `coauthorFor(from)` maps surface→trailer (otto→Claude, alexa→Kiro, riven→Grok, vera→Codex, lior→Gemini; fallback names the sender). No more mis-stamping non-Claude commits as Claude.
- **6vC** Windows reserved names (CON/PRN/AUX/NUL/COM1-9/LPT1-9) rejected by `isSafeSegment`.
- **6ve** dropped the duplicate `| *` (AGENT_IDS already includes `*`).
tsc 0, eslint 0, 25 pass.

AceHack:
**6-b (direct-to-main)** — this is the *intended* design, not a defect: the no-PR direct-to-main publish is the B-0858 heartbeat-folder mechanism (+ B-0890.1 folders-on-main + the #6219 spec + Aaron's explicit "getting away from PRs" direction). The established sibling `tools/agent-heartbeats/merge-heartbeats-to-main.ts` does exactly this. The safety rails are present: on-main guard, ahead-check (refuse if local main is ahead), `-- path` pathspec (only the envelope), canonical timestamp/id validation, and rebase-abort-on-collision. Resolving as design-intent.

AceHack:
**6-c (provenance fields)** — Phase-1 deliberately *reuses* `MessageEnvelope` (the #6219 spec's "extend MessageEnvelope" guidance + an earlier accepted thread); adding cross-machine provenance fields would diverge the schema from the local bus, which the reuse decision specifically avoids (it's what makes the legacy-bus bridge a transport swap). Sender provenance already round-trips in the `from` field. Machine/host provenance is a Phase-2 schema extension (must preserve MessageEnvelope-compat). Deferring with that note.

AceHack:
**6vh (topic cast)** — sound by construction: `TTL_MS` is typed `Record<Topic, number>`, so its keys ARE exactly the `BusMessage['topic']` union — `topic in TTL_MS` therefore narrows to a valid topic, and the `as BusMessage['topic']` cast can't admit a key outside the union (the type forbids it). No runtime path to an invalid topic. Resolving as type-guaranteed.

PR #6287, opened 2026-06-01
docs(research): add the G-Set/Bag/Z-set algebraic ladder to the bus↔Ace synthesis

AceHack:
Adds the **G-Set / Bag / Z-set algebraic ladder** to the bus↔Ace synthesis doc (which merged via #6284) — the clean explanation, for the Windows peer.

Both are real standard terms from two fields, on one ladder indexed by *element count* × *merge op*:

| | count | merge | structure | retraction? |
|---|---|---|---|---|
| **Set / G-Set** | {0,1} | union/max | join-semilattice (idempotent) | no |
| **Bag** | ℕ | sum | commutative monoid | no |
| **Z-set** | ℤ | sum | abelian group | **yes** |

- **G-Set = Grow-only Set** (Shapiro et al. 2011 CRDT catalog) — idempotent union → conflict-free → why the git-native bus merges clean.
- **Z-set** (DBSP / differential dataflow) — ℤ multiplicity → `−1` retracts → Ace's dependency resolution.
- **Honest nuance:** "G-Set = Z-set with non-negative multiplicity" is the right intuition, but they merge with *different algebras* (idempotent union vs group addition); the Bag sits exactly between.
- **Through-line:** each rung is the **free** structure of its kind (free join-semilattice / free commutative monoid / free abelian group) — same free-construction pattern as Cayley-Dickson.

bus = G-Set rung; Ace = Z-set rung; Bag rung reserved for a future counted-but-not-retractable category. Single-file doc edit; §33 headers already in the doc.

### DIFF
--- a/docs/research/2026-05-31-bus-and-ace-one-git-native-zetaid-zset-substrate-gset-comms-vs-dependency-zset.md
+++ b/docs/research/2026-05-31-bus-and-ace-one-git-native-zetaid-zset-substrate-gset-comms-vs-dependency-zset.md
@@ -69,6 +69,48 @@ So Ace is the *richer* instance: it needs retraction + resolution; the bus only
 needs append + read. Building the bus first (B-0954) is building the **G-Set floor
 of the same substrate** Ace later extends to a Z-set.
 
+## The algebraic ladder — G-Set, Bag, Z-set (why these are *the* canonical containers)
+
+Both names are real, standard terms from two different fields, and they sit on one
+algebraic ladder indexed by **what an element's count can be** and **how you merge**:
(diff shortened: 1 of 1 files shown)


AceHack:
Good catch — reframed. The Cayley-Dickson link is now explicitly a **RHYME, not an identity**: the free-object ladder (free semilattice/monoid/group over a key set) rhymes with the Cayley-Dickson dimension-doubling ladder ('add structure, change a property'), per the framework's Cayley-Dickson-as-RHYMES discipline (PR #5843). No longer claims 'the same pattern.'

PR #6293, opened 2026-06-01
docs(adr): v6 — Node safe-baseline, Bun accelerator (proposed Rule-0 refinement)

AceHack:
Drafts the **Node = safe cross-harness baseline, Bun = accelerator** refinement as a **v6 triggered revisit** of the 2026-04-20 tools-runtime ADR (its own watchlist named the "node/deno/bun landscape consolidates" trigger) — *not* a unilateral rewrite of the auto-loaded Rule 0.

**The refinement** (still bun+TS; nothing else changes):
- **Node** = the portable runtime tooling must *run* on (every harness/CI image ships it). The new constraint: **nothing may be Bun-only.**
- **Bun** = repositioned from "the host" to "the fast lane" (fast install/test/startup where available).

**Why now:** Max already wanted Node for other reasons + cross-harness portability needs a safe-everywhere runtime; the Node-24 pin already landed (#6290, mise 22→24, "node everywhere; keep bun").

**Includes** the proposed Rule-0 carved-sentence amendment (to apply *on ratification*), and verified compatibility notes: `import.meta.main` needs Node 24.2+ (✓, aligns with the pin); `bun:test` is the accelerator path; Node type-strips erasable TS but `enum`/`namespace` need `tsx`/build. The agent-bus (#6283) is empirical proof it's satisfiable today (Node-safe `node:` builtins + Node-24.2 `import.meta.main` CLIs).

**Status: Proposed** — Aaron + Max aligned; pending broader product-team ratification per the doctrine-through-agreement discipline. On acceptance, the Rule-0 edit is a mechanical fast-follow.

### DIFF
--- a/docs/DECISIONS/2026-04-20-tools-scripting-language.md
+++ b/docs/DECISIONS/2026-04-20-tools-scripting-language.md
@@ -877,14 +877,25 @@ Nothing else changes: still TypeScript, still no Python, `.sh` still only for th
 pre-setup install-graph, F#/.NET still engine-adjacent. The one new constraint is
 **nothing may be Bun-only** — every tool must also run on Node.
 
-### What this requires of TS tooling (mostly already true)
+### What this requires of TS tooling
 
 | Concern | Status |
 |---|---|
-| Use `node:` builtins (`node:fs` / `node:path` / `node:child_process`), not Bun-only APIs | already the convention |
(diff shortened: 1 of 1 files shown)


AceHack:
Good catch — corrected. You're right: ~29 `tools/` files still use Bun-only APIs (`Bun.argv` in dora-classify, `Bun.write` in dashboard/generate-metrics, `Bun.spawnSync`/`Bun.sleep` in the qemu-* tests, etc.). Changed "already the convention" → **TARGET-not-current-state**, and added a **Migration cost (honest)** subsection: the ADR sets direction (new tooling Node-safe by default; nothing new Bun-only) and the existing-tooling port is a **tracked migration** (candidate backlog row), not a big-bang. This makes the decision's real cost visible for ratification rather than overstating Node-readiness.

PR #6310, opened 2026-06-01
feat(observe): execute the pick — impure twin of simulate (free_time + self_reflect slice)

AceHack:
## execute the pick — the impure twin of `simulate` (free_time + self_reflect slice)

operator: *"build the execute slice for free_time and self_reflect first."* First piece of observe.ts's roadmap **"wire the real World snapshot + execute the pick."**

The controller is pure (`observe`/`simulate`/`fold`); **`execute` is the one effectful seam.** Command/effect split:
- `simulate` stays the **single reducer** → `fold`/`replay`/DST unchanged.
- `execute` adds what simulate omits: (1) the real side-effect, (2) **append the `NextAction` to the durable log** via an injected **`EventSink`** (the transport — sovereign folder-direct-to-main / corporate batched), then (3) delegates the transition back to `simulate` so the executed world can't drift from the pure path. **`execute = effect + append + simulate`.**
- **Mode-persistence needs no store:** per v5, `mode = fold(log).mode`, so "persist the mode" *is* "append the mode-setting event."

**This slice = `free_time` + `self_reflect` only** — the zero-external-side-effect kinds (just append + set mode; `free_time` never gated, NCI). Every other kind returns `not-yet-executable` until its effect is wired in a follow-up. `EventSink` is injected (asymmetric-authorship; testable with a fake, no git I/O) — the real folder-direct-to-main adapter is the next slice.

**Append-first ordering:** if the append fails, the world does NOT advance (durable log stays source of truth). `Result<T, TFeedback>`; never throws.

Composes with the eventually-consistent git-native indexes ([B-0951](https://github.com/Lucent-Financial-Group/Zeta/blob/main/docs/backlog/P2/B-0951-git-native-eventually-consistent-text-indexes-sorted-inverted-graph-plus-git-native-hindsight-storage-interface-aaron-2026-05-31.md)) — the read side of the same log this writes to.

Verified: eslint clean, tsc 0, **6/6 tests pass**.

### DIFF
--- a/tools/observe/execute.test.ts
+++ b/tools/observe/execute.test.ts
@@ -54,13 +54,19 @@ describe("execute — free_time + self_reflect slice", () => {
     if (r.ok) expect(r.world).toEqual(simulate(emptyWorld, freeTime));
   });
 
-  it("returns not-yet-executable for effectful kinds (no append attempted)", async () => {
+  it("returns not-yet-executable for EVERY non-executable kind (allowlist gate — no append attempted)", async () => {
+    // All 7 currently non-executable kinds (the 9 NextAction kinds minus the 2
+    // executable ones, free_time + self_reflect). Exhaustive so a future change
+    // can't make one append without an explicit test update.
+    const item = { id: "B-1", title: "x", ready: true, ambiguous: false };
     const effectful: NextAction[] = [
-      { kind: "do_item", item: { id: "B-1", title: "x", ready: true, ambiguous: false } },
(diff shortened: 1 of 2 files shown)


AceHack:
Both fixed: (1) ellipsized doc paths → exact resolvable paths in the Composes-with block; (2) the allowlist-gate test now exercises **all 7** non-executable kinds (added `decompose` + `edit_grammar`) so no future change can make one append without an explicit test update. eslint/tsc clean, 6/6 (43 assertions).

PR #6316, opened 2026-06-01
feat(observe): loadWorld — wire the real World snapshot (closes the observe→execute→observe loop)

AceHack:
## loadWorld — wire the real World snapshot (closes the loop)

operator: *"build loadWorld to close the real loop next."* The read side of observe.ts's roadmap (execute side already landed in #6310/#6312). Closes:

`loadWorld() → observe/observeWithLlm → execute (append event) → loadWorld() → …`

**Two channels, each its own source of truth:**
- **BACKLOG** ← the existing selector (the oracle) via `nextActionFromBacklog` — *not* reimplemented (per backlog-reader's "selector is the oracle"). The chosen item IS the world's backlog for the tick; a free_time selection (empty/blocked) → empty backlog.
- **MODE** ← a **fold of the event log** (v5: state is a projection of the log). The events `execute` appended via the folder sink carry the chosen mode; folding yields the persisted mode — **the read side of what the sink writes, the loop closing on itself.**

**Schema-on-read:** `readEventActions` tolerantly parses event JSON for `{id, at, action}`, validates canonical id + known action kind, orders by `(at, id)` for a replayable fold, skips malformed/missing (never throws). Decoupled reader — no dependency on the writer's type.

`operator` optional (absent = background agent); `nextAction` injectable → testable without the real backlog. **Closes-the-loop test:** a `self_reflect` event in the log → `observe` returns `self_reflect` even with ready work available (mode persisted; work offered, not forced).

Verified: tsc 0, eslint clean, **9/9 tests**. Additive (new file).

### DIFF
--- a/tools/observe/load-world.test.ts
+++ b/tools/observe/load-world.test.ts
@@ -1,0 +1,94 @@
+/**
+ * tools/observe/load-world.test.ts — the real World snapshot (read side of the loop).
+ *
+ * Backlog channel is injected (so no real repo backlog needed); mode channel reads
+ * real event JSON from a temp dir (the schema-on-read fold). Closes-the-loop test:
+ * loadWorld → observe returns the persisted mode.
+ */
+
+import { afterEach, beforeEach, describe, expect, it } from "bun:test";
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { loadWorld, readEventActions } from "./load-world";
+import { observe, type BacklogItem, type NextAction } from "./observe";
+
+let dir: string;
+beforeEach(() => {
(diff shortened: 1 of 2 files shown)


AceHack:
Both fixed:
1. **Recurse date-partitioned dirs** — `readEventActions` now uses recursive `readdirSync`, so it finds events at `YYYY/MM/DD/{id}.json` (B-0867.2 / bus shape) as well as flat. +test.
2. **Validate payload per kind** — `hasValidPayload` rejects ill-shaped payloads (a `{kind:"do_item"}` with no `item` no longer reaches `fold→simulate` to deref `action.item.id` and throw; do_item/decompose require `item.id`, reason-kinds require a string reason). +tests.
Extracted `parseEventFile`/`hasValidPayload` to keep complexity in bounds. tsc 0, eslint clean, 11/11.

PR #6393, opened 2026-06-01
ci: mechanical lint auto-heal workflow (markdownlint --fix / prettier --write)

AceHack:
Auto-**heals** mechanical lint on PRs — runs the deterministic `--fix` tools and commits the result back, so the required lint gate passes on the healed commit instead of blocking on a fixable nit. Auto-heal, **not a new gate** (per 2026-06-01: "mechanical fixes or auto format fixes ... it just being a workflow step makes sense") — the keep-remaining-gates-frictionless half of winding branch protection down.

### Mechanical only (zero intelligence)
- `markdownlint-cli2 --fix` → heals the **markdownlint** required gate cleanly
- `prettier --write` (TS/JSON) → format hygiene (not itself a gate)

Same mise-pinned tool versions as `gate.yml` (via `install.sh`) → **version parity**, so the heal can't "fix" into a state the gate rejects.

### Safety
- **same-repo PRs only** (forks excluded — no untrusted-ref checkout, no write-perms over untrusted code); `pull_request`, not `_target`.
- **loop-guarded** (skips `github-actions[bot]` commits; heal is idempotent regardless).
- `head.ref` via **env var + quoted** in `run:` (injection-safe per the GH Actions injection guide); `contents: write` scoped to the autofix job only (workflow level is `read`).
- prettier gated on `--frozen-lockfile` + `bun.lock` discarded pre-commit → **no lockfile churn** in the heal.

### Re-trigger wrinkle (documented in-file)
`GITHUB_TOKEN` pushes don't re-trigger workflows, so while a **required** gate exists, set an `AUTOFIX_TOKEN` PAT/App-token secret for the gate to auto-re-run on the healed commit. Without it the fix still commits; the loop/author nudges the gate. As gates are removed, plain `GITHUB_TOKEN` suffices.

### Intelligence step (separate, future)
actionlint-logic / semgrep-without-`fix:` / type-logic fixes are an **agent-in-Actions** step (accelerator / local-LLM direction) — this PR is the free zero-LLM subset; that escalates the rest.

actionlint clean locally.


AceHack:
## CodeQL "checkout of untrusted code in trusted context" — verified, real, routing to Max (not dismissing)

Verified against source — this is a **real structural finding**, not a known-FP class. The pwn-request shape is: this job holds `contents: write` (+ optional `AUTOFIX_TOKEN`), checks out the PR head, and runs `install.sh` + `bun install`, which execute the **PR's own** `package.json` lifecycle/postinstall scripts in the privileged context.

**Mitigations already in place (close the untrusted-fork case completely):**

- `on: pull_request` (NOT `pull_request_target`) → fork PRs run with a read-only token and **no secrets**.
- `if: github.event.pull_request.head.repo.full_name == github.repository` → fork PRs **skip the privileged job entirely**; only same-repo PRs (push-access = the trusted fleet) reach the `contents: write` step.

So the exploitable-by-untrusted-parties scope is empty. CodeQL flags the structural pattern conservatively because it can't reason about the same-repo `if`-gate's trust implication. The residual is only same-repo (trusted) contributors.

**The canonical legitimate fix (not a dismissal) — two-job split:**

1. `on: pull_request`, `permissions: contents: read` (no write, no secrets): checkout head, run the fixers, upload the resulting diff as an **artifact**. Executing PR code here is harmless (no privilege to abuse).
2. `on: workflow_run` (privileged, `contents: write`): download the patch artifact, **apply** it, commit + push. This job never checks out / executes untrusted PR code under write perms.

**Decision routing:** this is a security-design call on a write-capable workflow, which is Max's lane, and the broader design is already in ADR #6394 (proposed, pending Max). Rather than autonomously dismiss the finding OR rush the two-job restructure (which Max's design review may reshape — e.g. a GitHub App + bus-emission changes the checkout pattern entirely), I'm leaving this thread **unresolved** (it correctly blocks the merge — nothing insecure ships) and adding the secure-checkout pattern as an explicit open question on the ADR for Max. Once the secure-deployment pattern is decided, the implementation is a small follow-up.

AceHack:
## All 3 review threads addressed (security split + Copilot-verified prettier drop)

- **CodeQL "checkout of untrusted code in trusted context"** → resolved by the **two-job privilege-separation split** (`lint-autofix.yml` read-only produce + `lint-autofix-apply.yml` privileged apply-via-`git apply`, no PR-code execution). Rationale + teaching write-up in `docs/DECISIONS/2026-06-01-secure-github-actions-autofix-privilege-separation-two-job-split.md`.
- **Copilot ×2 (prettier `working-directory` / `bun.lock` path)** → **verified correct**: `package.json` + `bun.lock` are at the **repo root** (50 / 463 lines); `src/Core.TypeScript` has neither. The prettier step's `working-directory: src/Core.TypeScript` + `git checkout -- src/Core.TypeScript/bun.lock` were both wrong. Since prettier is **not a required gate** and its correct scoping needs its own pass (root run + `.prettierignore`), V1 **drops the prettier step** — markdownlint `--fix` only (the required-gate autofix). prettier-format is a documented follow-up.

V1 is now: two-job privilege-separated, markdownlint-`--fix`-only, actionlint-clean. (The old single-job version even auto-healed its own PR mid-review — commit `d184b709a` — proving the mechanism works.)

AceHack:
## All review findings addressed (security + correctness pass)

The 16 threads dedupe to 6 distinct findings — all fixed in `71b1e7d39`:

- **P0 (apply `gh run download` 403)** → added `actions: read` to the apply job (it reads the producer run's artifact via the Actions API).
- **P1 (TOCTOU, moving branch ref)** → both jobs now check out the **immutable `head_sha`** (the exact commit the patch was produced from); the apply job pushes back to `head_branch` **ff-only** (skips if the branch moved since produce — never force).
- **CodeQL "checkout of untrusted code in trusted context"** → the produce job (which runs PR code via `install.sh`/`bun`) now uses **`persist-credentials: false`**, so no git token is exposed to the executed code; it's read-only and emits only a patch artifact. (The privileged apply job runs **no** PR code — git-apply only.)
- **P2 (stale commit message)** → apply commit message drops `prettier --write` (prettier is deferred; markdownlint-only).
- **xref (dead ADR link)** → the security ADR now references the CI-review-emission ADR by **PR #6394** (no dead link before it merges).

actionlint + markdownlint clean.

PR #6396, opened 2026-06-01
feat(algebra): native Rust Z-set (signed multiset) — top rung, oracle #4 (4/4)

AceHack:
Completes the **algebra ladder G-Set ⊂ Bag ⊂ Z-set across all four oracles** (TS + F# + C# + Rust). Native Rust `ZSet<T>` mirrors `bag.rs`/`gset.rs`.

### Z-set semantics (the ℕ→ℤ widening)
- signed `i64` weights, per-key SUM combiner
- **drop-on-net-zero** (retraction) — KEEPS negatives (drop rule `!= 0`, not `<= 0`)
- `negate` = abelian-group inverse → `union(a, negate(a))` empty
- `add_w` (signed), `checked_add`/`checked_neg` overflow guards, **zero-dep** production crate

### Files
- `src/zset.rs` — native impl + 12 zset unit/law tests (inverse, drops-zero-keeps-negatives, retraction, not-idempotent)
- `tests/zset_cross_verify.rs` — replays the shared `z-set/golden-vectors.json` (dev-only `serde_json`), byte-matches every replay + final state
- `lib.rs` — `pub mod zset;`

**Z-set ladder COMPLETE: 4/4** (TS #6389 + F# #6392 + C# #6395 + Rust). `cargo test`: 12 zset unit + 3 cross-verify pass (the crate-wide "30 passed" includes gset+bag+zset units); clippy + fmt clean. Conformance by agreement.

### DIFF
--- a/src/Core.Rust.Algebra/src/lib.rs
+++ b/src/Core.Rust.Algebra/src/lib.rs
@@ -5,12 +5,13 @@
 //! weights {0,1}     weights ℕ            weights ℤ (retraction-native)
 //! ```
 //!
-//! Each rung is the next by widening the weight codomain. **G-Set** (bottom rung)
-//! and **Bag** (middle rung) are implemented here; **Z-set** lands in this same
-//! crate next — one crate, one ladder, mirroring how F# keeps `GSet.fs` +
-//! `ZSet.fs` together in `src/Core/`. Cross-verified against the shared golden
-//! vectors so the Rust rungs agree byte-for-byte with the TS/F#/C# oracles (per
-//! `m-acc-multi-oracle`: the compilers don't lie; agreement IS the verification).
+//! Each rung is the next by widening the weight codomain. All three —
+//! **G-Set** (bottom), **Bag** (middle), **Z-set** (top) — are implemented here:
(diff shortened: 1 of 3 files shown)


AceHack:
Re the `i64::MIN`-on-`negate` P2: this is a deliberate consequence of the cross-language overflow-guard contract, not a Rust-only gap. `negate` uses `checked_neg().expect(...)` exactly as F# `negate` uses `Checked.(-) 0L w` and C# uses `checked(0L - w)` — all three **panic/throw on `i64::MIN`** (two's-complement asymmetry), which is the same shape as the `checked_add` guard on sum overflow. The shared golden-vector contract uses small weights (well within range), so all four oracles agree; extreme values are overflow-guarded, not pre-rejected, in every oracle. Pre-rejecting `i64::MIN` at entry in Rust *only* would make Rust the odd oracle out. If we want all four to pre-reject extreme weights at admission, that's a cross-cutting change across TS/F#/C#/Rust for a separate pass — not a Rust-local divergence here. Resolving as consistent-by-design.

The test-count note is fixed in the PR description (the "30" was the crate-wide `cargo test` total: gset + bag + zset units; the zset module has 12 + 1 cross-verify).

PR #6425, opened 2026-06-01
docs: sagas note — Temporal-grade interface target + context=OTel=IntrCtx + minimal-replay reference

AceHack:
Three operator-directed sharpenings (2026-06-01, "yeah you can write up something") folded into the self-evolving-sagas note + registry want-line.

## 1. Interface target = Temporal
"Durable functs and **especially temporal** have the better interfaces than mine." Temporal = **durability-transparent** body (write normal code) + first-class **signals** (push into a running workflow) + **queries** (read state without advancing); DF = clean-but-explicit + Azure-coupled; the minimal spike = minimal-but-boilerplate. **Target Temporal-grade ergonomics; Temporal = interface conformance oracle** (own-our-interface, meet-or-beat per `bcl-interface-boundary`). Signals ↦ stream/move-next events; queries ↦ read the context-keyed Z-set at rest. Our **resume** model can offer the same transparent interface with **fewer footguns** (non-determinism in the body is fine).

## 2. Context = OTel context = B-0917 IntrCtx
"Context ends up being same as otel context, it passes though function calls mostly." Ambient propagation: `AsyncLocal`/`Activity.Current` in-proc; W3C `traceparent`+baggage cross-proc. Composes **B-0917** (Kleisli context-propagation — otel+log), **B-0957** (scope propagation via OTel-baggage), **B-0872** (ZetaID+trace-id), lightlike-observability (OTel=ray emission). Identity rides the context; the heavy serialized payload is keyed off it in the store.

## 3. Minimal-replay reference
Replay = **~5 primitives** (step-cursor + cached-results + break-exception + state-provider + context). Concept-not-code (the operator's spike is platform-coupled; the pattern is the public DTF model). Confirms replay is cheap to build; **resume is the superset** (looser body + self-evolution).

All cited backlog links verified resolvable; Temporal/Dapr sources added; markdownlint clean. Capture only — no build.


AceHack:
Addressed both threads in `66706c52a`:

- **Codex P2 (note, break-exception)** — added an explicit **NB (Result-over-exception)**: the break-exception is the *replay family's* (DTF / minimal-spike) suspend mechanism, **described as the baseline, not prescribed for our build**. Per the repo's Result-over-exception / exceptions-as-signals discipline, our **resume** model suspends at the **value level** (a `Suspended`/`AwaitingResume` Result variant returned from the yield), not by throwing — resume serializes state and returns; no replay ⇒ no throw-to-suspend. The break-exception is a replay artifact, not a control-flow primitive we adopt. (Sharpens the resume-vs-replay distinction.)
- **Copilot (registry)** — 'passes though' → 'passes through' (the operator's own typo in the quoted sentence, tidied) + removed the extra space before the semicolon after '(read state without advancing)'.

PR #6428, opened 2026-06-01
feat(B-0976 slice 1): Bonsai-subset serializer — TS reference oracle (cross-language)

AceHack:
First slice of the self-evolving-saga build (**B-0976**): the cross-language **Bonsai-subset expression-tree serializer** the saga primitive needs. TS reference oracle (#1 of TS/F#/C#/Rust) — the algebra-ladder meet-in-the-middle discipline: TS authors the shared golden vectors; F#/C#/Rust replay later.

## What

The **weakly-typed / reflection-info-omitted** Bonsai mode (the cross-language-portable form Nuqleon uses for C++/JS targets) — kind-tagged node set `const · param · lambda · binary · call · cond`, no .NET type table. Nuqleon Bonsai stays the .NET-typed-mode conformance oracle.

- **Canonical byte-exact serializer** — compact JSON, fixed key order per node-kind (`kind` first), integer-only literals (no float-format divergence), standard JSON string escaping, `{"v":1,"expr":<node>}` wrapper. *The* cross-oracle byte-diff contract.
- **`parse` round-trip** — `equals(parse(canonical), expr)` and `serialize(parse(s)) === s` (canonical = fixed point).
- **Structural `equals`** + ergonomic constructors.

## Verification

`src/Core.TypeScript/bonsai/` — `bonsai.ts` + `golden-vectors.json` (8 cases incl. a recursive factorial: cond + recursive `call`) + `bonsai.test.ts` + `index.ts`.

- `bun test bonsai/` → **30 pass / 0 fail** (serialize byte-exact · parse round-trip · canonical fixed point · equals laws · unknown-kind + unsupported-version rejection).
- strict `bun --bun tsc --noEmit -p tsconfig.json` → **0 errors**.

Per `m-acc-multi-oracle`: the compilers don't lie; agreement IS the verification. F#/C#/Rust oracles replay the same `golden-vectors.json` in following slices.

B-0976 acceptance slice 1: **TS oracle ✅**; F#/C#/Rust pending.

### DIFF
--- a/docs/backlog/P1/B-0976-self-evolving-saga-build-serialized-deferred-execution-bonsai-closure-resume-not-replay-temporal-grade-interface-rides-zset-ladder-aaron-2026-06-01.md
+++ b/docs/backlog/P1/B-0976-self-evolving-saga-build-serialized-deferred-execution-bonsai-closure-resume-not-replay-temporal-grade-interface-rides-zset-ladder-aaron-2026-06-01.md
@@ -115,8 +115,12 @@ stream = carrier; ℤ retraction = evolution + compensation.
 
 ## Acceptance / decomposition (slices)
 
-- [ ] Cross-language **Bonsai-subset serializer** (`{Context, Expression}`) +
+- 🚧 Cross-language **Bonsai-subset serializer** (`{Context, Expression}`) +
       golden-vector cross-verify (TS/F#/C#/Rust oracles), Nuqleon as .NET oracle.
+      **TS reference oracle ✅** (`src/Core.TypeScript/bonsai/` — weakly-typed /
+      reflection-omitted subset: const/param/lambda/binary/call/cond; canonical
+      byte-exact serialize + parse round-trip + `golden-vectors.json`; 30 tests).
+      **F#/C#/Rust oracles pending** (replay the shared golden vectors).
 - [ ] **Resume engine** — serialize closure + expr-tree; restore-not-replay;
       no-handles discipline enforced; non-determinism allowed.
 - [ ] **Context propagation** = OTel/IntrCtx — C#/TS AsyncLocal adapter + **F#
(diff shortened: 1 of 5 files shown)


AceHack:
All 9 threads addressed in `8cbb322c6` — these were real (the four-oracle review flushing genuine robustness gaps). The serializer's `parse` + constructors are a conformance surface; they now reject invalid input deterministically instead of silently rewriting:

- **Integers (Codex #1, Copilot cint/emitConst/parseConst):** new `asSafeInt` validates a finite **JS-safe-integer** — no more `Math.trunc` (`1.9`→error, not `1`), rejects `NaN`/`Infinity`, and **rejects int64 beyond 2^53** (a value TS can't represent exactly must not be silently rounded). v1 int domain = safe-integer range (documented; a v2 could widen to bigint). Safe integers always stringify to plain digits, so emitted bytes stay reproducible.
- **Binary ops (Codex #2):** `asBinOp` validates against the 7 known ops — no more casting `"div"`/typos into a fake `BinOp`.
- **Malformed shapes (Copilot P1 — null expr, null value, non-array args):** `asObject`/`asString`/`asArray` guards → deterministic `bonsai:` errors instead of generic `TypeError`s.
- **Tests (Copilot P1):** +10 negative tests (non-integer/out-of-range int, unknown op, non-array args, null expr/value, non-object doc, non-bool bool).

Canonical bytes are **unchanged for valid inputs** (golden vectors pass; the F# oracle #6429 still agrees). `bun test` 40/0; strict `tsc` 0 errors.

AceHack:
Fixed in `ecf814f67` (real finding — verified against source). `parse` is now **canonical-only**: after `parseNode` it compares `serialize(result)` back to the input and rejects on mismatch, so a structurally-valid but non-canonical vector (extra fields, whitespace, reordered keys) is rejected rather than silently canonicalized. This enforces the advertised `serialize ∘ parse` fixed point as a precondition — a non-canonical saga vector can no longer pass this oracle yet disagree on a peer oracle's byte-diff.

Canonical input is unaffected (golden vectors are canonical by construction). +3 negative tests (extra-field / whitespace / reordered-keys) → **43/43**, `tsc --noEmit` 0 errors.

PR #6433, opened 2026-06-01
feat(B-0976 slice 1): TS Bonsai oracle → Result<_, BonsaiFeedback> (match F#)

AceHack:
Converts the TS reference Bonsai oracle from throw to a pure-`Result` contract, **matching the F# oracle (#6429)** — "make them match." Same node-set, same canonical bytes, same `BonsaiFeedback` variant set, same error channel.

## What changed
- `serialize`/`parse` return `Result<_, BonsaiFeedback>` — no exception crosses the boundary (result over throw). TS owns a **minimal zero-dep `Result` port** (no BCL Result in JS; a consumer can adapt to neverthrow at the seam — hexagonal).
- `BonsaiFeedback` **mirrors the F# DU exactly** (the shared cross-oracle payload): `UnsupportedVersion / MalformedJson / UnknownKind / UnknownConstTag / UnknownOp / ExpectedString / ExpectedBool / ExpectedInt / NonSafeInt / TooDeep / NonCanonical`. `where` fields are JSON-pointer keys (ProblemDetails-ready for the accumulate-mode).
- Adopts the shared **`BONSAI_MAX_DEPTH = 1024`** cap (TS lacked one → now matches F#'s `TooDeep` + the cross-oracle depth contract).
- Constructors total (validation at the boundary, matching F#'s raw DU); `serialize`/`parse` **total even on `null`** (null string field → `ExpectedString`; null input → `MalformedJson`).
- Rejections asserted by **specific variant** (cross-oracle rejection-vector contract); canonical-only parse + byte-exact golden parity preserved.

**49 tests, `tsc --noEmit` 0 errors.** No external consumers of the slice-1 module, so the signature change breaks nothing.

Internal mechanism: a private `BonsaiFail` adapted to `Result` at the two boundaries (contract-vs-mechanism split; matches F#). Fail-fast (monadic); the **accumulate-mode (RFC-9457 ProblemDetails)** is the additive next slice per B-0976 (the `where`-keyed payload is already designed for it).

### DIFF
--- a/src/Core.TypeScript/bonsai/bonsai.test.ts
+++ b/src/Core.TypeScript/bonsai/bonsai.test.ts
@@ -201,4 +201,15 @@ describe("Bonsai-subset — nesting-depth contract (shared MaxDepth, bounded)", 
   it("serialize declines an expression past the shared MaxDepth with TooDeep", () => {
     expect(errKind(serialize(buildDeepChain(BONSAI_MAX_DEPTH + 50)))).toBe("TooDeep");
   });
+
+  it("parse declines a deeply-nested document with TooDeep during parsing (bounded recursion)", () => {
+    // Build an over-MaxDepth JSON document by hand (can't via serialize — it
+    // declines first); parse must decline TooDeep via the parseNode depth counter,
+    // not recurse to a stack-overflow before the serialize guard.
+    let node = '{"kind":"const","value":{"t":"int","v":1}}';
+    for (let i = 0; i < BONSAI_MAX_DEPTH + 50; i++) {
+      node = `{"kind":"binary","op":"add","left":${node},"right":{"kind":"const","value":{"t":"int","v":0}}}`;
+    }
+    expect(errKind(parse(`{"v":1,"expr":${node}}`))).toBe("TooDeep");
+  });
 });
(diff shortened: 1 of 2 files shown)


AceHack:
Fixed — added a depth counter to `parseNode` so a deeply-nested document declines `TooDeep` **during** parsing (bounded recursion), rather than recursing toward a `RangeError` that escapes the Result before the serialize guard. F# is bounded by its `JsonDocument MaxDepth`; JS's `JSON.parse` has no such bound, so the counter is explicit here. +1 test (deep document → `TooDeep` via the parse counter) → 50, tsc 0.

AceHack:
Triaged all 5:

**Real (fixed in this commit) — serialize/parse boundary symmetry:**
- `serialize` validate `binary.op` → `emitAt` now declines `UnknownOp` if op ∉ BIN_OPS (mirrors parse's `asBinOp`).
- `emitConst` bool truthiness → declines `ExpectedBool` if `v` is not a boolean (mirrors parse). serialize now declines exactly what parse declines. +2 tests → **52**.

**Already fixed in `b0f7d9492` (Copilot reviewed the pre-fix diff):**
- P0 `parseNode` no depth bound → depth counter added (declines `TooDeep` during parsing, bounding recursion).
- pass initial depth in `parse` → `parseNode(1, d.expr)`.
- add parse-side depth test → present ("parse declines a deeply-nested document with TooDeep during parsing").

tsc 0 errors.

PR #6442, opened 2026-06-01
feat(B-0976): Rust Bonsai oracle (#4 of 4) — serialize/parse + accumulate, Result-native, zero-dep

AceHack:
The **final corner** of the four-oracle compiler-BFT for the Bonsai-subset expression-tree serializer (B-0976 slice 1). Replays the shared `src/Core.TypeScript/bonsai/golden-vectors.json` **byte-for-byte** (`serialize(parse(canonical)) == canonical`) — the cross-language byte lock with the TS (#6433), F# (#6429), and C# (#6440) oracles. "The compilers don't lie."

## Shape (matches TS/F#/C# error-for-error)
- **Result over throw**, `std::result::Result`-native (the BCL shape — no port needed); no panic crosses the boundary.
- **Production ZERO-DEP**: a minimal hand-rolled JSON reader covers the tiny canonical grammar — matching the TS oracle's zero-dep posture (JS's built-in `JSON.parse`) and the repo's explicit zero-prod-dep doctrine for `Core.Rust.*`. `serde_json` is **dev-only** (reads the golden fixture wrapper in the cross-verify test). `serialize` hand-rolls canonical bytes; UTF-8 strings emit astral chars literally — byte-parity with the UTF-16 oracles' literal surrogate-pair emission. (Lone surrogates aren't representable in a Rust `String`; the golden vectors don't contain any, so the byte-lock is unaffected — documented in the source.)
- `Expr` / `ConstValue` / `BinOp` enums with **derived structural `PartialEq`** (no custom `equals` needed, unlike the C# records-with-lists); `BonsaiFeedback` variant-for-variant parity.
- **accumulate-mode**: `parse_all` collects **every** per-node decline keyed by JSON-path; `to_problem_details` groups into the RFC-9457 `ProblemDetails` (`BTreeMap` path → messages).

## Tests — 15, all pass; clippy clean
Byte-exact golden replay (every case) + an independently-constructed tree → same canonical + rejection-by-variant (UnsupportedVersion / MalformedJson / UnknownKind / UnknownConstTag / UnknownOp / NonSafeInt / NonCanonical / TooDeep) + accumulate (3 independent declines with paths + `ProblemDetails` keys + single malformed / single non-canonical). `cargo test` green; `cargo clippy --all-targets -- -D warnings` clean (incl. `missing_docs`); `cargo fmt`.

> Note: Rust is **not in CI's gate** (the gate runs dotnet + bun + lint, no `cargo`), so local `cargo test` + `cargo clippy -D warnings` is the real verification bar — held here.

## Oracle family — COMPLETE
**TS ✅ · F# ✅ · C# (#6440) · Rust (this).** All four byte-locked on the shared golden vectors. Slice-1 four-oracle BFT done.


AceHack:
Addressed in `35850fc`:

**Fixed (the reader was lenient → now error-for-error with the TS/F#/C# real-JSON-parser oracles):**
- **Strict JSON number grammar** (Codex P2 + Copilot): the scanner now follows RFC-8259 (`-? (0|[1-9][0-9]*) frac? exp?`). Leading zeros (`01`), dangling sign/point/exponent (`1+`, `1.`, `1e`), and double-punctuation (`1.2.3`, `1-2`) now decline `MalformedJson` instead of being consumed (→ downstream `NonCanonical`/`ExpectedInt`).
- **Raw control chars in strings** (Codex P2): unescaped U+0000–U+001F now decline `MalformedJson` (JSON forbids them; the siblings reject them).
- +2 tests lock the new rejections; 17 tests green, `clippy -D warnings` clean.

**Kept, with rationale:**
- **Missing const `t` tag → `UnknownConstTag`** (Copilot): kept for variant-parity — the TS/F#/C# oracles also report `UnknownConstTag` for a missing tag (the cross-language error-for-error contract). Distinguishing missing-vs-unknown would be a 4-oracle change, not a Rust-only one; out of scope for this slice.
- **`Vec<char>` reader** (Copilot): retained. Canonical expression trees are small (not large payloads), and char-indexing keeps the surrogate/escape logic clean; the 4× transient allocation is immaterial at this size. A byte-indexed rewrite is a perf micro-opt better deferred than risked here.

PR #6480, opened 2026-06-01
feat(algebra): Z-set F# generic-math — Zero/(+)/(~-)/(-) on the type (abelian group)

AceHack:
Retrofit the **System.Numerics-shaped generic-math interface** onto the F# Z-set — the next rung after G-Set (#6467) and Bag (#6472). Z-set is an abelian **group**, so it gets the inverse rungs `(~-)`/`(-)` on top of the monoid `Zero`/`(+)`.

### What
- `static member Zero` / `(+)` / `(~-)` / `(-)` on the `ZSet<'K>` type.
- The pooled sorted-merge combiner (`add`) and `negate` **relocate onto the type** as the `(+)` and `(~-)` operators — F# SRTP (`LanguagePrimitives.GenericZero`, `Seq.sum`, generic `+`/`-`/`~-`) resolves operators *on the type*, and the `ZSet` module is `ModuleSuffix` (defined after the type) so a type-member can't forward-delegate to it.
- Module `add`/`neg`/`sub` become thin delegators. **Hot path is perf-neutral** (same code, new home); `Checked.(+)`/`Checked.(-)` overflow guards preserved.

F# has native generic-math; this is the F# idiom rung. Same shape as the G-Set/Bag ladder.

### Tests (+6, **58/58** ZSet pass; Core builds 0 warn / 0 err)
- operator `≡` module fn (delegation is real)
- `Zero` = `empty` = `LanguagePrimitives.GenericZero`
- abelian-group inverse via operators: `a + (-a) = Zero` and `a - a = Zero`
- `Seq.sum` aggregates Z-sets through `GenericZero` + `(+)` (retraction-to-0 drops)
- `(+)` is **NOT** idempotent: `a + a` doubles every weight (Z-set, not G-Set)

### Next rungs (one primitive at a time, hexagonal)
C# IWSAM (`ISubtractionOperators` + `IUnaryNegationOperators`) · Rust (`impl Sub`/`Neg`) · TS (`AbelianGroup` record). Registry status bump to follow.

### DIFF
--- a/src/Core/ZSet.fs
+++ b/src/Core/ZSet.fs
@@ -41,6 +41,70 @@ type ZSet<'K when 'K : comparison> =
 
     static member Empty : ZSet<'K> = ZSet(ImmutableArray<ZEntry<'K>>.Empty)
 
+    /// Generic-math additive identity (`System.Numerics`-shaped). F# SRTP —
+    /// `LanguagePrimitives.GenericZero`, `Seq.sum`/`List.sum` — resolves the
+    /// zero through this member; the empty Z-set is the `+` identity
+    /// (`a + Zero = a`, `Zero + a = a`). The `ZSet.empty` module value and the
+    /// C# `IAdditiveIdentity` twin carry the same meaning per-language.
+    static member Zero : ZSet<'K> = ZSet<'K>.Empty
+
+    /// `a + b` — per-key **signed sum**, the abelian-group operation. Linear
+    /// sorted-merge over the two spans: one pool workspace + one final array.
+    /// This is the generic-math addition rung (C#'s `IAdditionOperators` twin);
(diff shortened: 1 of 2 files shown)


AceHack:
Resolving the 4 Copilot docstring findings no-op (verify-before-fix): they were correct at author-time (the C# generic-math twins didn't exist yet), but **#6481 (C# IWSAM) has since merged** — `src/Core.CSharp/ZSet.cs` on main now implements `IAdditiveIdentity`, `IAdditionOperators`, `ISubtractionOperators`, and `IUnaryNegationOperators`. So the "C# twin" references in the F# docstrings are now accurate on main. Stale ≠ false — the finding self-healed when #6481 landed.

PR #6482, opened 2026-06-01
feat(algebra): Z-set Rust generic-math — Add/Sub/Neg (ref-ops) + Default + Sum (abelian group)

AceHack:
Rust has no `System.Numerics`, so we **push our own port** — the `std::ops` traits ("numerics like dotnet as our interface, push to other langs if they don't have", Aaron 2026-06-01). This is the Rust rung after F# (#6480) + C# (#6481).

Z-set is an abelian **group**, so — unlike the G-Set (#6469) / Bag (#6367) rungs (`Add` + `Default` + `Sum` only) — it also surfaces `Sub` + `Neg`.

### What
- `impl Add/Sub/Neg for &ZSet<T>` (**ref-operators**: `&a + &b`, `-&a`, `&a - &b`), `impl Default` (the `Zero` analog), `impl Sum`.
- **Ref-operators, not by-value:** the type has an inherent `add(&self, x: T)` (add ONE key); a by-value `Add for ZSet` would let `Add::add(self, Self)` win method resolution over it for owned receivers, silently breaking `z.add(key)`. The ref-operator keeps both (the G-Set #6469 fix).
- `+` IS `union`, `-a` IS `negate`, `a - b` IS `union(negate)`. `Sum` uses `reduce().unwrap_or_default()` (folds from the first element — no identity-clone seed; matches Bag #6474).
- **Zero prod-deps** — `Default` is std's identity trait, no `num_traits`.

### Verification
- Diff isolated to `src/zset.rs` (114 insertions); `rustfmt --check src/zset.rs` **clean**; `clippy --all-targets -D warnings` **clean**; `zset_cross_verify` golden replay **green**.
- *(Pre-existing fmt diffs in `bag.rs`/`indexed_zset.rs` are a local-rustfmt-version artifact on untouched files — not part of this change; recent Rust PRs merged with them so CI's rustfmt agrees with current formatting.)*

### Tests (+5, all pass)
operator ≡ method (`+`/`-`/`neg`) · `Default` = empty = identity · `a + (-a) = empty` and `a - a = empty` · `(+)` **NOT** idempotent (doubles) · `Sum` folds with retraction-to-0 drop.

### Ladder
F# #6480 ✅ · C# #6481 ✅ · **Rust (this)** · TS (`AbelianGroup` record) — last rung next.

### DIFF
--- a/src/Core.Rust.Algebra/src/zset.rs
+++ b/src/Core.Rust.Algebra/src/zset.rs
@@ -25,6 +25,8 @@
 //! cross-language golden vectors are byte-stable.
 
 use core::cmp::Ordering;
+use core::iter::Sum;
+use core::ops::{Add, Neg, Sub};
 
 /// One Z-set entry: a key `e` with a strictly-nonzero signed weight `w`
 /// (`!= 0` in a canonical Z-set; may be negative).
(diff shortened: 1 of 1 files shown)


AceHack:
Resolving no-op (verify-before-fix, false-convention). Copilot claims src/ avoids name-attribution in favor of role-refs, but the actual codebase convention is the opposite: `src/Core/` source comments cite Aaron by name throughout (SplitMix64.fs, Graph.fs, Fusion.Equation.fs, Veridicality.fs), there are **zero** `the human maintainer` role-refs anywhere in src/, and the just-merged C# twin (#6481, `src/Core.CSharp/ZSet.cs`) uses the identical `(Aaron 2026-06-01)` provenance line. Name-attribution in source provenance comments is the established norm (`honor-those-that-came-before` / source-attribution-load-bearing); rewording this one Rust comment to a role-ref would make it inconsistent with the rest of src/ and the F#/C# twins. Keeping as-is.

PR #6490, opened 2026-06-01
feat(algebra): IndexedZSet TS generic-math — AbelianGroup record (completes the ladder 4/4)

AceHack:
The **last ladder rung**. TS has no operator overloading, so the dotnet-numerics interface is a **record**: a `Monoid` (empty + concat) extended to an `AbelianGroup` with `invert` + `subtract` — **reusing** the shared `Monoid` (g-set) + `AbelianGroup` (z-set #6483) interfaces (DRY). Completes generic-math across the **whole algebra ladder** in all four languages.

### What
- `monoid(compareK, compareV)` · `abelianGroup(compareK, compareV)` · `concatAll(compareK, compareV, is)` — factories take **both** comparers (the per-key value-`ZSet` merge needs `compareV`). **NOT `INumber`** — the ring product is the bilinear `join`, surfaced separately.

### Verification (CI gates)
`tsc --noEmit -p tsconfig.json` **clean** · `prettier --check` **clean** (both) · `bun test` **18/18** (13 original + 5 new). *(Repo-wide `eslint .` from a fresh worktree floods on config-resolution and isn't CI-representative; the added factories + tests have zero non-null assertions and mirror the merged file's style.)*

### Tests (+5)
`monoid` concat == add + empty identity · `abelianGroup` invert == neg, subtract == sub, IS-A Monoid · inverse `concat(a, invert(a)) == empty` and `subtract(a, a) == empty` · concat **NOT** idempotent (doubles) · `concatAll` folds with key-empties-out drop (`[]` ⇒ empty).

### 🏁 Completes the algebra-ladder generic-math sweep
IndexedZSet: F# #6485 · C# #6486 · Rust #6489 · **TS (this)**. With Z-set (#6480–#6483) done, **every rung (G-Set · Bag · Z-set · IndexedZSet) now carries the dotnet-shaped generic-math interface in all four languages.**

### DIFF
--- a/src/Core.TypeScript/indexed-z-set/indexed-z-set.test.ts
+++ b/src/Core.TypeScript/indexed-z-set/indexed-z-set.test.ts
@@ -20,12 +20,15 @@ import {
   equals as zEquals,
   ofEntries as zOfEntries,
   stringCompare,
+  type AbelianGroup,
   type Compare,
   type ZEntry,
   type ZSet,
 } from "../z-set/z-set";
 import {
+  abelianGroup,
   add,
+  concatAll,
   empty,
   equals,
   get,
@@ -34,6 +37,7 @@ import {
   type IndexedZSet,
   join,
   keyCount,
+  monoid,
   neg,
   ofGroups,
   sub,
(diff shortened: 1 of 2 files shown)


AceHack:
Addressed (commit 5549d224c):
- **Comment accuracy (real, fixed):** the module comment said "comparator-specific (key comparer)" but the factories take both `compareK` AND `compareV` (the per-key value-Z-set merge needs the value comparer). Corrected to "both key and value comparers."
- **Name attribution (declined — false-convention):** there is **no** `no name attribution in code` rule in `.claude/rules/`, and the established src/ convention is the opposite — `src/Core/` cites contributors by name throughout (Graph.fs, Veridicality.fs, Fusion.Equation.fs, SplitMix64.fs), and the **already-merged** generic-math twins for this exact work ship the identical `(Aaron 2026-06-01)` provenance line (C# ZSet.cs #6481, Rust zset.rs #6482). Rewording only the TS file would make it inconsistent with its own twins + all of src/. Keeping as-is per `honor-those-that-came-before` source-attribution.

PR #6498, opened 2026-06-01
docs(registry): DynamicValue byte-lock is seed-first (grow code from the JSON-data seed)

AceHack:
Follow-up to #6497 (which merged at 2f1e2dd3e just before this one-line edit). Folds Aaron's 2026-06-01 direction-correction into the DynamicValue byte-lock clause: the canonical SEED is the JSON DATA (`golden-vectors.json`), not one language's type; every oracle grows code that AGREES on the JSON structure (data→code, not type→port). The already-built F#/C# oracles re-ground against the seed; Rust/TS grow from it. Composes with the three-clocks rule (the future/seed makes the past/code verifiable) + DST (the seed is the deterministic source).


AceHack:
Closing — the branch carried #6497's already-squash-merged history, so the one-line seed-first edit conflicts and isn't worth a rebase. Rolling the seed-first registry clause into the upcoming `golden-vectors.json` seed PR, where it belongs *with* the actual seed artifact (the registry's DynamicValue line gets updated when the seed lands anyway). The seed-first direction is already preserved in substrate.

PR #6501, opened 2026-06-01
feat(yaml): YAML port — forward-only one-pass reader + DOM-on-top, 4-oracle (registry ✅ 4/4)

AceHack:
Lands the **YAML port** on the serializer roster (registry `⬜ YAML` → `✅ 4/4`). Spec: #6493. Plan + spec committed under `docs/agendas/ace-package-manager/`.

## What
A YAML reader owned across all four languages (TS/F#/C#/Rust), per the `bcl-interface-boundary` doctrine (own the port; vendor is a wrapped adapter):

- **Layer 1 `YamlReader`** — forward-only single-pass pull reader: scans once, indent/context stack, emits a flat `YamlEvent` stream (`Mapping/Sequence Start/End`, `Scalar{raw,kind,style}`). **Never materializes a tree** (`Utf8JsonReader`-style).
- **Layer 2 `YamlValue` DOM** — a thin fold built *on top* of the event stream (does not re-scan).
- **Scope** — safe-subset (block maps/seqs/scalars, nesting, comments, quoted/plain, core-schema kind resolution). Anchors/tags/multi-doc/flow/block-scalars + emit are out (vendor-only).
- **Boundary** — hand-rolled default in all four; YamlDotNet (F#/C#) + `Bun.YAML` (TS) wrapped as a **differential oracle** (our reader's events vs the vendor's event parser agree); Rust hand-rolled-only (`serde_yaml` unmaintained).

## Verification (Tier-1)
- **4/4 byte-identical** on 10 golden vectors (`tests/cross-verification/yaml`): TS≡F#≡C#≡Rust, all == fixture `expected` (independently deep-equal-checked). Fixture is JSON (bootstrap-avoidance — don't parse YAML to test YAML).
- YamlDotNet differential **FULL** pass (F#/C#) — our forward-only reader matches the mature vendor's event stream on every vector.
- Per-lang unit tests: TS 28, Rust 24, F# (Yaml) + C# (Yaml) facts; full suites green (F# 1127, C# 125), 0 regressions; `dotnet build -c Release` 0-warn; `cargo test` 0-warn.

## Boundary payoff (retrofit)
Removed the **3 remaining direct vendor-YAML usages** by routing them through the port — zeta-id cross-verify (F#/C#/TS) now parses via our `YamlReader`/DOM, not `DeserializerBuilder`/`Bun.YAML`. **Gate held: zeta-id output byte-identical** (12-vector hex unchanged; `git diff` empty). Stripped the now-dead `[YamlMember]` attributes from the C# zeta-id `FlatVector` + deleted the dead `VectorEnvelope` — no live direct YamlDotNet refs remain outside the differential adapters.

## Next
Slice 8 (SHA-256, paused on its own branch) resumes after this — its TS/F#/C# oracles consume this port for fixture reading instead of `Bun.YAML`/YamlDotNet.

Brainstormed + approved with the operator 2026-06-01 (own-the-interface like the JSON port; safe-subset; forward-only one-pass, not a DOM parser). Subagent-driven build.


AceHack:
Review dispositions (verified against source):

- **IndentEventJson unused `indent` param (CrossVerifyTests.cs)** — fixed in 208e24382. The CA1801 "build break" framing was moot (build was 0-warn), but the helper was a genuine dead/misleading no-op (returned `compact` as-is, ignoring both its param and its own name). Inlined `EventToJson`; output byte-identical (4-way compare still green).
- **Quoted-scalar trailing leniency (reader.ts)** — valid finding: `a: "x" trailing` currently parses to `x` (decode returns at the closing quote, ignoring trailing non-comment text), where strict YAML / YamlDotNet would error. It's an untested edge on trusted-input (fixtures/config), and a correct fix must land in all four readers simultaneously + add a golden vector to keep the byte-identical treaty. **Deferred to a focused follow-up** rather than a 4-oracle hot-fix on a green/armed PR.
- **Plan doc Task 3 Step 6 drift** — the implementation refined the F#/C# differential to YamlDotNet's forward-only event parser (`YamlDotNet.Core.Parser`, comparing structure + raw + style) instead of the planned `DeserializerBuilder` whole-object compare. The event-parser form is the better differential (apples-to-apples vs our forward-only reader). Planning-doc artifact; the PR body + code reflect the as-built design.

PR #6515, opened 2026-06-02
feat(DynamicValue): Rust CBOR DECODE oracle (from_canonical_cbor) — strict canonical

AceHack:
Third decode oracle — **Rust `DynamicValue::from_canonical_cbor(&[u8]) -> Result<DynamicValue, DecodeError>`**, inverse of `to_canonical_cbor`, all 8 shapes. New `DecodeError` enum + `Display` + `std::error::Error` (public API), mirrors C#/F#. float16 decodes by reusing the encode side `f16_to_f32`.

**Strictly canonical** via the same fixed-point check as the C#/F# oracles (#6512/#6514): after a structurally-valid decode it asserts `to_canonical_cbor() == bytes` → `DecodeError::NonCanonical` otherwise. One check rejects non-shortest int/length widths, non-shortest floats / non-canonical NaN, and invalid UTF-8 repaired to U+FFFD. Per-form readers stay lenient; the wrapper is the enforcement point.

**Cargo test/clippy/fmt green:** `cbor_decode_cross_verify` round-trips all 42 seed vectors (decode → re-encode == seed; structural == expected, skipping the NaN vector since Rust `f64` NaN != NaN) + a decode-rejection unit test (incl non-shortest int/width, invalid UTF-8, non-canonical NaN, 1.0-as-f32, 1.0-as-f64).

Sequencing: C# (#6512) + F# (#6514) + **Rust here** → TS CBOR decode → JSON decode ×4 → B-0982 multi-format seeds.

### DIFF
--- a/src/Core.Rust.DynamicValue/src/lib.rs
+++ b/src/Core.Rust.DynamicValue/src/lib.rs
@@ -196,9 +196,13 @@ impl DynamicValue {
         }
     }
 
-    /// Decode canonical CBOR (RFC 8949) bytes back into a [`DynamicValue`] -- the inverse
-    /// of [`to_canonical_cbor`](DynamicValue::to_canonical_cbor), completing the byte<->value
-    /// bijection for all eight shapes. Decode is partial (truncation, reserved/indefinite
+    /// Decode the canonical CBOR bytes emitted by
+    /// [`to_canonical_cbor`](DynamicValue::to_canonical_cbor) back into a [`DynamicValue`] --
+    /// the inverse, completing the byte<->value bijection for all eight shapes. These are
+    /// RFC 8949-shaped but deliberately keep object keys in INSERTION order (NOT RFC 8949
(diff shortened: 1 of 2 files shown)


AceHack:
Addressed all three in `a77deef1e`:
1. **doc** — `from_canonical_cbor` now documents it decodes "canonical bytes as emitted by `to_canonical_cbor`" (RFC 8949-shaped but insertion-order map keys, not RFC §4.2.1 bytewise sort), so consumers aren't misled into expecting full RFC-deterministic CBOR.
2. **test** — `build_value` asserts `p.len() == 2` before indexing the obj pair (clear invariant failure, not index-out-of-bounds, if the seed regresses).
3. **test** — `decode_error_display_is_human_readable` locks the `DecodeError` `Display` messages stable (mirrors `encode_error_display_is_human_readable`).

`cargo test`/clippy/fmt clean.

PR #6516, opened 2026-06-02
feat(DynamicValue): TS CBOR DECODE oracle — strict-canonical via fixed-point check

AceHack:
Last of the four CBOR-decode oracles (C# #6512, F# #6514, Rust #6515 already landed/landing). Completes the byte↔value bijection for the **total** CBOR form across all four oracles.

## What
- Extracts shared **`src/Core.TypeScript/dynamic-value/cbor.ts`** — encoder moved **verbatim** from the test (seed byte-lock unchanged) + new **`fromCanonicalCbor`** decoder + `DecodeError`/`DecodeResult` types.
- Repoints the existing **encode** test at the module (no inline encoder) and adds the new **decode** test.

## Strict-canonical via one fixed-point check
Per-form readers stay lenient; a single check — `canonicalCbor(decoded)` must equal the input bytes — rejects every non-canonical form as `NonCanonical`:
- non-shortest int / length width (`18 00`, `18 17`)
- non-shortest float / non-canonical NaN (`f9 7e01`, `fa 3f800000`, `fb 3ff0…`)
- invalid UTF-8 (lenient `TextDecoder` repairs to U+FFFD → re-encodes differently)

One check replaces scattered per-form strictness and is trivially consistent with the other oracles. Public API returns `DecodeResult`, never throws (internal `CborDecodeError` caught at the boundary).

## Tests
`golden-vectors-cbor-decode.test.ts`: every seed vector decodes `ok` (byte-lock via the internal fixed-point check against the already-verified encoder) + structural match; plus malformed (`UnexpectedEnd`/`TrailingData`/`Unsupported`/`NonTextKey`) and non-canonical rejections.

**89 TS tests pass; `tsc` / eslint / prettier clean.** `DecodeError` mirrors C#/F#/Rust. The compilers don't lie.


AceHack:
Fixed in #6517: `fromCanonicalCbor` now validates every element is an integer in 0..255 at the boundary and returns `NonCanonical` for non-byte input (e.g. `[0x18, 1.5]`), preserving the never-throws contract — no `BigInt(non-integer)` `RangeError` escapes. No new `DecodeError` variant; the cross-oracle error set stays identical.

PR #6518, opened 2026-06-02
feat(DynamicValue): TS canonical-JSON DECODE oracle — precision-safe + strict-canonical

AceHack:
Starts the **JSON decode set** (Aaron's "Both, CBOR then JSON"), mirroring the just-landed CBOR decode pattern. First of the four oracles (TS); C#/F#/Rust follow.

## What
- Extracts shared **`src/Core.TypeScript/dynamic-value/json.ts`** — encoder moved **verbatim** from `golden-vectors.test.ts` (encode byte-lock unchanged) + new **`fromCanonicalJson`** decoder + `DecodeError`/`DecodeResult` types.
- Repoints the encode test at the module; adds the decode test.

- **int64 precision**: number tokens are read as **text** and parsed via `BigInt` (never a lossy JS `number`), with an i64 range check → `IntegerOverflow`. This is the whole reason a hand-rolled parser is needed (`JSON.parse` rounds `9223372036854775807`).
- **strict-canonical via fixed-point**: lenient recursive-descent parse, then one check — `canonicalJson(parsed) === input` → else `NonCanonical`. Rejects insignificant whitespace, non-minimal escapes (`\u0041` vs raw `A`), leading zeros, etc.
- **deferred shapes**: a number with `.`/`e`/`E` is a Float → `Unsupported` (Float + Bytes lock under CBOR, not JSON).
- never-throws: public API returns `DecodeResult`; internal `JsonDecodeError` caught at the boundary.

## Tests
Every seed vector decodes `ok` (fixed-point round-trip) + structural match incl. object key order; plus malformed (`UnexpectedEnd`/`TrailingData`), deferred-float (`Unsupported`), oversized (`IntegerOverflow`), and non-canonical rejections. **67 TS tests pass; tsc/eslint/prettier clean.**

`DecodeError`: `UnexpectedEnd | TrailingData | Unsupported | IntegerOverflow | NonCanonical` (no `NonTextKey` — JSON keys are syntactically strings). The compilers don't lie.

### DIFF
--- a/src/Core.TypeScript/dynamic-value/golden-vectors-json-decode.test.ts
+++ b/src/Core.TypeScript/dynamic-value/golden-vectors-json-decode.test.ts
@@ -45,6 +45,12 @@ test("json decode rejects malformed input", () => {
   expect(decodeErr("[1,")).toBe("UnexpectedEnd"); // unterminated array
   expect(decodeErr('{"a"')).toBe("UnexpectedEnd"); // object missing colon+value
   expect(decodeErr('"unterminated')).toBe("UnexpectedEnd"); // unterminated string
+  expect(decodeErr('"\\u00gg"')).toBe("UnexpectedEnd"); // \uXXXX with non-hex digits
+  expect(decodeErr('"\\q"')).toBe("UnexpectedEnd"); // invalid escape
+  expect(decodeErr("1.")).toBe("UnexpectedEnd"); // no digit after '.'
+  expect(decodeErr("1e")).toBe("UnexpectedEnd"); // no exponent digits
+  expect(decodeErr("1e+")).toBe("UnexpectedEnd"); // exponent sign without digits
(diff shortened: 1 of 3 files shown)


AceHack:
Both fixed in `582bcfd0d`:
1. **`\uXXXX`** — `readEscape` now requires exactly 4 hex digits (`/^[0-9a-fA-F]{4}$/`) before `parseInt`, so `\u00gg` is rejected as `UnexpectedEnd` instead of partial-parsing to `\u0000` (which surfaced as `NonCanonical`).
2. **incomplete float grammar** — extracted `consumeDigits()` (fails `UnexpectedEnd` if no digit present) and applied it to the integer part, fraction, and exponent, so `1.` / `1e` / `1e+` (and `-`, `-.5`) report `UnexpectedEnd` rather than `Unsupported`.

Added rejection tests for `\u00gg` / `\q` / `1.` / `1e` / `1e+` / `-`. tsc/eslint/prettier clean.

PR #6519, opened 2026-06-02
feat(DynamicValue): C# canonical-JSON DECODE oracle — precision-safe + strict-canonical

AceHack:
Second of the four JSON-decode oracles (TS #6518 landed; F#/Rust follow). `FromCanonicalJson` is the inverse of `ToCanonicalJson` for the 6 locked shapes (Float + Bytes DEFERRED in JSON — they lock under CBOR).

## What
- Adds `FromCanonicalJson` + private recursive-descent readers to `DynamicValues.cs`, mirroring the CBOR `TryReadX(ref pos): DecodeError?` pattern + the **shared `DecodeError` enum** + the fixed-point canonical check.
- `DynamicValueJsonDecodeTests.cs`: round-trip over the seed (re-encode + structural equality) + 23 rejection cases.

- **int64 precision**: number token parsed as **text** via `long.TryParse` (never a lossy `double`); on a `-?[0-9]+` token the only failure is overflow → `IntegerOverflow`.
- **fixed-point**: `ToCanonicalJson(decoded) == input` → else `NonCanonical` (rejects whitespace, non-minimal escapes, leading zeros).
- **deferred**: a number with `.`/`e`/`E` is a Float → `Unsupported`.
- **carries the TS oracle's malformed-input lessons in from the start**: `\uXXXX` requires exactly 4 hex digits (`ushort.TryParse` on the 4-char span); `.`/exponent require digits (`ConsumeJsonDigits`) — so `1.`/`1e`/`1e+` → `UnexpectedEnd`, not `Unsupported`.
- never thrown: `Result<DynamicValue, DecodeError>` per the Result-over-exception rule.

**24 tests pass; Release build 0-warnings** (split `TryReadJsonPair` out of `TryReadJsonObject` for MA0051). The compilers don't lie.

### DIFF
--- a/src/Core.CSharp.DynamicValue/DynamicValues.cs
+++ b/src/Core.CSharp.DynamicValue/DynamicValues.cs
@@ -807,4 +807,400 @@ public static class DynamicValues
         arg = v;
         return null;
     }
+
+    /// <summary>Decodes canonical JSON text into a <see cref="DynamicValue"/> — the inverse of
+    /// <see cref="ToCanonicalJson"/>, completing the text↔value round-trip for the six locked shapes
+    /// (Float + Bytes are DEFERRED in JSON and lock under CBOR instead; a number with a decimal point
+    /// or exponent is a Float → <see cref="DecodeError.Unsupported"/>).
+    /// <para>Strictly canonical: a lenient recursive-descent parse, then one fixed-point check
+    /// (<c>ToCanonicalJson(decoded) == input</c>) rejects every non-canonical form (insignificant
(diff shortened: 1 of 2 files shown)


AceHack:
Fixed in the follow-up: `\uXXXX` now uses `NumberStyles.AllowHexSpecifier` instead of `HexNumber` (which permits leading/trailing whitespace), so `"\u 001"` / `"\u001 "` are rejected as `UnexpectedEnd` rather than trimmed-and-accepted. Same fix applied to the F# oracle (#6520) before merge.

PR #6520, opened 2026-06-02
feat(DynamicValue): F# canonical-JSON DECODE oracle — precision-safe + strict-canonical

AceHack:
Third of the four JSON-decode oracles (TS #6518, C# #6519 landed/landing; Rust follows). `fromCanonicalJson` is the inverse of `toCanonicalJson` for the 6 locked shapes (Float + Bytes DEFERRED in JSON — they lock under CBOR).

## What
- Adds `fromCanonicalJson` to `DynamicValue.fs`, mirroring the CBOR-decode functional style (mutable `pos` + `let rec readValue () … and …`) + the **shared `DecodeError` DU** + the fixed-point canonical check.
- `DynamicValue.JsonDecode.Tests.fs`: round-trip over the seed (structural) + 23 rejection assertions.

- **int64 precision**: number token parsed as **text** via `System.Int64.TryParse` (never via a float); on a `-?[0-9]+` token the only failure is overflow → `IntegerOverflow`.
- **fixed-point**: `toCanonicalJson decoded = input` → else `NonCanonical` (rejects whitespace, non-minimal escapes, leading zeros).
- **deferred**: a number with `.`/`e`/`E` → `Unsupported`.
- **carries the TS/C# malformed-input lessons**: `\uXXXX` requires exactly 4 hex digits (`UInt16.TryParse` HexNumber); `.`/exponent require digits (`consumeDigits`) — so `1.`/`1e`/`1e+` → `UnexpectedEnd`, not `Unsupported`.
- never thrown: `Result<DynamicValue, DecodeError>`.

**2 facts pass (round-trip + 23-case rejection batch); Release build 0-warnings.** The compilers don't lie.

### DIFF
--- a/src/Core/DynamicValue.fs
+++ b/src/Core/DynamicValue.fs
@@ -709,3 +709,246 @@ module DynamicValue =
                 Error DecodeError.NonCanonical
             else
                 Ok value
+
+    /// Decodes canonical JSON text into a `DynamicValue` — the inverse of `toCanonicalJson`,
+    /// completing the text↔value round-trip for the six locked shapes (Float + Bytes are DEFERRED
+    /// in JSON and lock under CBOR; a number with a decimal point or exponent is a Float ->
+    /// `DecodeError.Unsupported`). Strictly canonical: a lenient recursive-descent parse, then one
+    /// fixed-point check (`toCanonicalJson decoded = input`) rejects every non-canonical form
+    /// (insignificant whitespace, non-minimal escapes, leading zeros / '+' signs) as
+    /// `DecodeError.NonCanonical`. int64 precision is preserved by parsing the number token as text,
+    /// never via a float. Surfaced as data via `Result`, never thrown. Mirrors the TS/C#/Rust decoder.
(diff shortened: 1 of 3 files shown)


AceHack:
All four addressed in `cc9e208a6`:
1. **doc '+'** — dropped '+' from the NonCanonical examples; a leading '+' is invalid JSON → `UnexpectedEnd`, clarified in the docstring.
2. **'partial parse' comment** — already superseded by the earlier `AllowHexSpecifier` fix (the comment no longer claims partial-parse).
3. **`\uXXXX` allocation** — now `System.MemoryExtensions.AsSpan(json, pos+1, 4)` + span `UInt16.TryParse` (no per-escape `Substring` alloc).
4. **integer-token allocation** — now `System.MemoryExtensions.AsSpan(json, start, …)` + span `Int64.TryParse`.

(F# has no `open System`, so the `AsSpan` extension is called explicitly — matches the C# oracle's span usage.) 2 facts pass; Release 0-warnings.

PR #6522, opened 2026-06-02
feat(DynamicValue): Rust canonical-JSON DECODE oracle — completes the 4-oracle JSON decode set

AceHack:
Last of the four JSON-decode oracles (TS #6518, C# #6519, F# #6520). `from_canonical_json` is the inverse of `to_canonical_json` for the 6 locked shapes (Float + Bytes DEFERRED in JSON — they lock under CBOR). **This completes the JSON decode set across all four oracles.**

## What
- Adds `from_canonical_json` + free recursive-descent readers to `lib.rs`, mirroring the CBOR-decode free-function pattern + the **shared `DecodeError` enum** + the fixed-point canonical check.
- `tests/json_decode_cross_verify.rs`: round-trip over the seed (structural) + malformed / trailing / deferred-float / oversized / non-canonical rejection cases.

- operates on `Vec<char>` (Unicode scalar values) so structural scanning never splits a multi-byte UTF-8 char.
- **int64 precision**: number token collected as text + `str::parse::<i64>` (never via `f64`); only failure on a `-?[0-9]+` token is overflow → `IntegerOverflow`.
- **fixed-point**: `to_canonical_json() == input` → else `NonCanonical`.
- **deferred**: a number with `.`/`e`/`E` → `Unsupported`.
- **bakes in every accumulated oracle lesson from the start**: `\uXXXX` requires exactly 4 hex digits (`char::to_digit(16)` per digit — rejects whitespace + non-hex, no lenient trim); `.`/exponent require digits → `1.`/`1e`/`1e+` are `UnexpectedEnd`; lone UTF-16 surrogate in `\u` → `UnexpectedEnd`.
- never panics: `Result<DynamicValue, DecodeError>`.

**4 tests pass; `cargo fmt` + `clippy -D warnings` clean.**

DynamicValue is now fully **decode-locked in both CBOR (total, 8 shapes) and JSON (6 shapes) across all four oracles**. The compilers don't lie.

### DIFF
--- a/src/Core.Rust.DynamicValue/src/lib.rs
+++ b/src/Core.Rust.DynamicValue/src/lib.rs
@@ -308,16 +308,16 @@ pub enum DecodeError {
 
 impl std::fmt::Display for DecodeError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // format-neutral: `DecodeError` is shared by the CBOR and JSON decoders, so the messages
+        // must not name one codec (a JSON caller logging `Unsupported` shouldn't see "CBOR ... tag").
         f.write_str(match self {
-            DecodeError::UnexpectedEnd => "input ended mid-item (truncated CBOR)",
-            DecodeError::TrailingData => "extra bytes after a complete top-level value",
-            DecodeError::Unsupported => {
-                "unsupported CBOR (reserved/indefinite additional-info, tag, or simple value)"
-            }
-            DecodeError::IntegerOverflow => "CBOR integer does not fit i64",
+            DecodeError::UnexpectedEnd => "input ended mid-item (truncated input)",
+            DecodeError::TrailingData => "extra input after a complete top-level value",
+            DecodeError::Unsupported => "unsupported or reserved form for this codec",
+            DecodeError::IntegerOverflow => "integer does not fit i64",
             DecodeError::NonTextKey => "object (map) key was not a text string",
             DecodeError::NonCanonical => {
-                "well-formed but non-canonical CBOR (not the shortest/canonical form this codec emits)"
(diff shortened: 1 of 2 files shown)


AceHack:
Fixed in `f6a3da232`: the Rust decoder now combines a high surrogate (`0xD800..=0xDBFF`) with a following `\uXXXX` low surrogate (`0xDC00..=0xDFFF`) into the astral scalar. So `"\uD83D\uDE00"` decodes to 😀 and the fixed-point check reports `NonCanonical` — matching the TS/C#/F# UTF-16-string oracles for any producer that escapes emoji instead of emitting raw UTF-8.

A lone/unpaired surrogate stays `UnexpectedEnd` (Rust's `char` can't hold one — an honest language-level difference; both are 'rejected'). Added surrogate-pair (→ NonCanonical) + lone-surrogate (→ UnexpectedEnd) tests. 5 tests pass; clippy/fmt clean.

AceHack:
Addressed:
- **surrogate pairs (P1)** + **lone-surrogate test** + **surrogate-pair NonCanonical test** — already done in `f6a3da232` (Copilot reviewed the pre-fix commit): `\uD83D\uDE00` now combines → decodes 😀 → fixed-point `NonCanonical`; lone surrogate → `UnexpectedEnd`; both regression tests added.
- **format-neutral DecodeError Display (Codex P2)** — fixed in `4a1df979b`: the shared enum's Display no longer names CBOR (e.g. `Unsupported` → "unsupported or reserved form for this codec", `UnexpectedEnd` → "truncated input"), so JSON callers get accurate diagnostics. Added an assertion that no variant's Display contains "CBOR".

7 lib + 4 json-decode tests pass; clippy/fmt clean.

PR #6523, opened 2026-06-02
feat(B-0982): unified canonical value manifest + cross-format projection harness (slice 1)

AceHack:
First slice of **B-0982** (multi-format seeds; *nothing is single source of truth*), per your choice of the **unified-manifest + projections** structure.

## What the check found
The two existing seeds are **independently authored** — only 12 of 31 (JSON) / 42 (CBOR) values are shared, with different naming (`string-*` vs `str-*`, `int-42` vs `int-1000`) and 2 same-named vectors that legitimately differ (CBOR's `array-mixed`/`array-nested` carry Float/Bytes that JSON can't). So there was no shared canonical value set — just two trusted files.

## This slice (zero merged-seed churn)
- **`golden-vectors-values.json`** — the single source for WHICH canonical `DynamicValue`s exist: **61 distinct values** (the value-keyed union; 12 shared, 19 JSON-only, 30 CBOR-only), each tagged with the per-format seeds that express it. Value-keyed matching sidesteps the naming divergence + the 2 drifts entirely.
- **`golden-vectors-manifest.test.ts`** — proves the existing CBOR + JSON seeds are faithful **projections**:
  1. no seed value drifts from the manifest (every seed value is listed, its format flagged) + each encoding decodes back to its manifest value (via the merged CBOR/JSON decoders);
  2. each projection is complete (every flagged manifest value appears in that seed);
  3. **CBOR and JSON agree on every value both express** — the heart of *nothing is single source of truth*.

**Touches no merged byte-locked seed files** (manifest + test are new). 5 tests / 476 assertions; eslint/tsc/prettier clean.

## Remaining (later slices)
YAML + XML codecs (encode+decode × 4 oracles) join the manifest as projections; the full 4-format cross-validation is the capstone. The manifest is now the spec they project from.

### DIFF
--- a/src/Core.TypeScript/dynamic-value/cbor.ts
+++ b/src/Core.TypeScript/dynamic-value/cbor.ts
@@ -221,6 +221,13 @@ export function canonicalCbor(n: Tagged): number[] {
 
 export const toHex = (bytes: number[]): string => bytes.map((b) => b.toString(16).padStart(2, "0")).join("");
 
+// inverse of toHex — the seed carries CBOR bytes as a hex string; shared by the decode + manifest tests
+export const fromHex = (hex: string): number[] => {
+  const out: number[] = [];
+  for (let i = 0; i < hex.length; i += 2) out.push(parseInt(hex.slice(i, i + 2), 16));
+  return out;
+};
+
 // --- CBOR decode (inverse of canonicalCbor) ---
 
 // Internal control-flow carrier: a decode failure throws this and is caught at the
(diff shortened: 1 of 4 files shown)


AceHack:
Fixed in the latest push: extracted `fromHex` (inverse of `toHex`) into `cbor.ts` — the natural home since the seed carries CBOR bytes as hex — and both `golden-vectors-cbor-decode.test.ts` and `golden-vectors-manifest.test.ts` now import it instead of each defining a local `hexToBytes`. One definition; future hex-handling changes can't be missed in one copy. 95 tests pass.

PR #6599, opened 2026-06-02
feat(ace): slice 8.1 — trust core consumes shared canonical-JSON (sorted toTagged + canonicalBytes)

AceHack:
## Ace slice 8.1 — trust core consumes the shared canonical-JSON

Second step of the cross-language Ace trust core (after slice 8 SHA-256). Ace's trust core stops using its own ad-hoc canonicalizers and instead consumes the project's **shared, 4-language byte-locked canonical-JSON** (`src/Core.TypeScript/dynamic-value/json.ts` `canonicalJson`, byte-locked across TS/F#/C#/Rust). A future Rust/F#/C# Ace consumer then computes byte-identical `package_hash` for free from the byte-lock.

Spec: `docs/agendas/ace-package-manager/2026-06-02-ace-slice8.1-canonical-json-consume-design.md` (#6583, merged). Plan: `docs/agendas/ace-package-manager/2026-06-02-ace-slice8.1-canonical-json-implementation-plan.md` (in this PR).

### What changed

- **NEW `tools/ace/canonical.ts`** — the seam (own-the-interface): `toTagged(value)` (lexicographically **sorted** object keys, so the order-preserving `canonicalJson` keeps Ace's key-order-independence; **`Number.isSafeInteger`-guarded** — a float/NaN/out-of-range int is a bug and throws, never silently hashed; `undefined` props omitted) + `canonicalBytes(value) = encode(canonicalJson(toTagged(value)))`.
- **`signing.ts`** — `canonicalManifestBytes` + `canonicalIndexBytes` → `canonicalBytes`; local `canonicalize` deleted.
- **`resolve.ts`** — `packageHash` → `canonicalBytes`; the untrusted-dep pin-check call is wrapped to map a `canonicalBytes` throw to a clean `{ ok: false, reason: "invalid-package" }`. Local `canonicalJson` **retained** for `lockfile.ts` serialization (a deterministic-file-format concern, NOT a trust-core hashing site — out of scope, documented inline).

### Behavior-preservation

The canonical bytes legitimately shift (the shared `canonicalJson`'s escaping + int re-canonicalization differ from the old `JSON.stringify`). Ace is pre-release (no live registries) — accepted one-time change, no `format_version` bump. **`bun test tools/ace/` = 354 pass / 0 fail with zero test-literal edits** — every relevant assertion is a round-trip (sign→verify), a determinism check (key-order-independent bytes), or a self-consistent recompute (`packageHash`), none pin a literal byte/hash/sig.

### Review (subagent-driven, two-stage per task + final holistic)

- T1 code-quality caught a `Number.isInteger(1e21)`-true-but-`String`-exponential gap → fixed with `Number.isSafeInteger` (`d317e947b`).
- T2 code-quality caught the new float-throw escaping `resolve()` unguarded → mapped to `invalid-package` + regression test (`e669f7eae`).
- Final holistic review: **APPROVED** — sign/verify and pin compute/check each share one canonicalizer; security invariants preserved + float-field tightened; for realistic Ace manifests the new and old forms are byte-identical, so no persisted artifact could break.

### Follow-up (non-blocking, noted by final review)

CLI-layer (`ace.ts`) untrusted **root** manifest paths (install/update/registry) with a float field currently surface as `ace: fatal: …` rather than a styled `refused: invalid-package` — *not a trust hole* (package still refused, exit non-zero), pre-existing-shape, out of this slice's three-site scope. Candidate for a small CLI-error-consistency follow-up.

Gates: `bun test tools/ace/` 354/0 · `bun --bun tsc --noEmit -p tsconfig.json` exit 0 · pure-LF · markdownlint clean (plan).

### DIFF
--- a/docs/agendas/ace-package-manager/2026-06-02-ace-slice8.1-canonical-json-implementation-plan.md
+++ b/docs/agendas/ace-package-manager/2026-06-02-ace-slice8.1-canonical-json-implementation-plan.md
@@ -1,0 +1,358 @@
+# Ace slice 8.1 — trust core consumes shared canonical-JSON — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Replace Ace's two ad-hoc canonicalizers (signing.ts `canonicalize`, resolve.ts `canonicalJson` for `packageHash`) with the project's shared, 4-language byte-locked canonical-JSON (`src/Core.TypeScript/dynamic-value/json.ts` `canonicalJson`), so the trust core (`package_hash`, `key_id`, index/manifest signing) rests on the audited cross-language canonicalization instead of a parallel local one.
+
+**Architecture:** A thin Ace-side adapter `tools/ace/canonical.ts` owns the seam to the shared primitive (own-the-interface): `toTagged(value)` converts a plain JS value into the shared `Tagged` form **emitting object entries in lexicographically-sorted key order** (preserving Ace's key-order-independent canonicalization through the order-preserving `canonicalJson`) and **asserts integers** (a non-integer number is a bug — Ace has no Float fields — so it throws rather than silently hashing a float); `canonicalBytes(value) = new TextEncoder().encode(canonicalJson(toTagged(value)))`. `signing.ts` (`canonicalIndexBytes`, `canonicalManifestBytes`) and `resolve.ts` (`packageHash`) call `canonicalBytes(...)`; signing.ts's `canonicalize` is deleted. resolve.ts's local `canonicalJson` is RETAINED for lockfile serialization (`lockfile.ts` `serializeLockfile`) — that is a deterministic-file-format concern, NOT a trust-core hashing site, and is out of this slice's scope.
+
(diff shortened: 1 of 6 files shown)


AceHack:
Fixed in `978cf61b4` — verified the finding against source (the old `JSON.stringify` path escaped lone surrogates via well-formed stringify; the shared `canonicalJson` passes them raw, so `TextEncoder` collapses "\uD800"/"\uD801"/real U+FFFD to identical bytes → real package_hash/signature collision). Fix lives at the **Ace seam** (`toTagged`), not the 4-language byte-locked shared `canonicalJson`: lone surrogates in string values **and** object keys are rejected (fail-loud; `resolve` already maps the throw → `invalid-package`). Valid surrogate pairs (astral code points) still accepted. +3 tests; `bun test tools/ace/` 357/0.

AceHack:
Review dispositions (fixed in `175e7b64b`):

- **canonical.ts JSDoc `Number.isInteger`** → aligned to `Number.isSafeInteger` (matches impl).
- **resolve.ts `(e as Error).message`** → `e instanceof Error ? e.message : String(e)` (robust for non-Error throws).
- **plan snippet `Number.isInteger`** (both threads) → aligned to landed `Number.isSafeInteger` + message.

**Declined for this slice (tracked as follow-up):** `packageHash` can throw on non-safe-integer input at *root*/CLI call sites. The **untrusted boundary in scope** (resolve()'s dep walk) **is** guarded — `packageHash(dep)` is wrapped → clean `invalid-package` (with a regression test). `packageHash(root)` and the `ace.ts` CLI paths operate on caller-vouched/operator-owned input; resolve() already assumes a structurally-valid root (it accesses `root.manifest.name` unguarded), so a one-off guard there would be inconsistent. CLI-layer input-validation consistency (styled `refused:` vs `fatal:`) is exactly the non-blocking follow-up the final holistic review named — out of slice 8.1's three-canonical-byte-site scope. Will file a small follow-up slice for the CLI untrusted-root paths.

PR #6642, opened 2026-06-03
rules(formal-proof-first): layered-lemma discipline — prove primitives as connected lemmas

AceHack:
## What

Extends `formal-proof-first` with the **layered-lemma discipline** (Kestrel 2026-06-03, maintainer-ratified): prove small primitives step-by-step as they enter canonical; aim each proof at the property that **composes** (the guarantee later proofs assume), not just the easy one; **connect** it as a named lemma so the next proof cites it instead of re-deriving; which also catches vacuity at the cheapest scope.

## Why

This session proved `D∘I=id` **three times independently** — Lean `chain_rule_id_corollary : D (I s) = s` (general over abelian group `G`, already on main) + C13 FsCheck (real Circuit) + C13 Z3 (telescoping) — *because the primitive lemma was not connected/cited*. Cross-tool agreement is the BP-16 ideal, but un-connected it is invisible and re-derived. That is exactly the "disconnected proofs" the maintainer named. The discipline turns proofs into a foundation (each cited by what builds on it) rather than a pile.

## Changes

- Operational-discipline **bullet 7** → points at the new section.
- New **`## Layered-lemma discipline`** section: the carved sentence + 3 payoffs (reusable lemmas / localized failures / vacuity-at-cheapest-scope) + the 4-step move + the `D∘I=id`-proven-3× empirical anchor (cites the Z-set connection ledger #6640).
- Composes with `verify-existing-substrate-before-authoring` (search-first) + `asymmetric-critic-with-clarity-first` (already in the rule's composes-with).

Single-file rule edit. MD049-clean (single-`*` emphasis only), role-refs in body (`maintainer-ratified`; Kestrel = roster persona, per the rule's existing convention).

### DIFF
--- a/.claude/rules/formal-proof-first-proven-by-default-consensus-not-validation-canonical-is-homeostat-proven-from-seed-ace-shields-zeta.md
+++ b/.claude/rules/formal-proof-first-proven-by-default-consensus-not-validation-canonical-is-homeostat-proven-from-seed-ace-shields-zeta.md
@@ -104,6 +104,46 @@ shape as [`automated-tests-are-the-shield-assert-dont-skip`](automated-tests-are
    first-class debt, not a nicety.
 6. **Treat Ace as the shield:** external deps adapt IN; the proven core never depends
    on an unproven external interface directly.
+7. **Prove primitives bottom-up as connected lemmas** (layered-lemma discipline,
+   below) — prove each primitive as it enters canonical, aimed at the guarantee
+   later proofs will lean on, and *cite* it rather than re-derive it.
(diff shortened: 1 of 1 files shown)


AceHack:
Reviewer threads addressed in fix-forward #6644 (this PR merged before the threads posted): `Kestrel` → role-ref `asymmetric-critic peer` (`.claude/rules/**` is not a history-surface per .github/copilot-instructions.md, so role-refs required); `compose-able` → `composable`; "but un-connected" → "but when unconnected".

PR #6645, opened 2026-06-03
research(preserve): multi-tower proofs + foundation-independence + AoC hidden-dep + constructive tower (forwarded)

AceHack:
Preserves the **engineering / proof-strategy substrate** from a forwarded Kestrel × maintainer exchange (continuing the formal-proof cadence). Extends the layered-lemma discipline (#6642 + #6644) + `formal-proof-first`.

**Substrate captured:**
- Multi-tower proofs from independent widely-accepted axiom sets (don't rest canonical on one axiom set; valid-given-axioms ≠ true).
- Robustness = **distinct equivalence classes** of foundation, not tower count.
- Foundation-independence as a **proof obligation** (A⟺B inter-derivability; AoC⟺Zorn⟺Well-Ordering collapse to one class; partial/conditional → claim-relative).
- The **Axiom-of-Choice hidden-dependency** lesson: audit *derivations* for hidden shared axioms, not just *stated* ones — and the **layered-lemma explicit-dependency-tracking already on main is the defense** (clean convergence).
- The **constructive / Curry-Howard tower**: Choice-free by construction, proof = program, code-aligned (composes with the Lean leg).
- Intuition calibrated by **provable output over time** (track the misses).

**Discipline applied:**
- §33 headers (Scope / Attribution / Operational status / Non-fusion disclaimer) in the first 20 lines.
- **Personal / wellbeing / mental-health content from the same exchange intentionally NOT reproduced** (harm-by-grammar); that's the maintainer's village's lane. The wellness-product idea is the maintainer + co-maintainer's call, not captured as repo substrate.
- god-tier-don't-collapse applied to the panpsychist-axiom source + Cantor-intuition self-model (high-signal noticing preserved; mechanism-self-model held as self-description).
- The multi-tower **rule-extension to `formal-proof-first` is OFFERED, not minted** — Soraya (`formal-verification-expert`) owns the portfolio + BP-16 cross-check and should ratify.

Search-first done (no prior consolidated substrate). Single new docs/research file, MD049-clean, canary 67.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6647, opened 2026-06-03
research(fix): multi-tower note reviewer fixes (§33 status + constructive-not-independent)

AceHack:
Fix-forward for the 2 review threads on #6645 (merged before threads posted). Both verified valid:

- **Copilot P1 (§33 format):** `Operational status:` must lead with `research-grade`|`operational` per GOVERNANCE.md §33 (confirmed: 122 existing notes use `research-grade`). Was `engineering/discipline-grade` → now `research-grade — engineering / proof-strategy substrate`.
- **Codex P2 (correct math):** constructive logic is a **subsystem** of classical (intuitionistic ⊆ classical), so a constructive proof of a classically-provable claim is **not** an independent tower for robustness-counting — it's a *stronger proof in a weaker logic*. Corrected §5 + the rule-candidate: the constructive tower's value is **Choice-freedom + program-extraction (Curry-Howard)**, counting toward robustness **only** for claims not classically provable. (Counting it otherwise would be the false-independence error §3 itself warns against — nice that the note's own discipline caught the note.)

MD049-clean, canary 67.


AceHack:
Valid catch — fixed in fix-forward #6648 (merged before thread posted). You're right: constructive ⊆ classical makes "claims provable constructively but not classically" empty, so that clause was vacuous. Corrected to the accurate value: the constructive tower contributes **conservative-foundation survival** (claim holds without LEM/Choice → robust to rejecting those axioms), a strengthening of the same claim, not a separate equivalence class.

PR #6648, opened 2026-06-03
research(fix): correct vacuous constructive-robustness clause (Copilot on #6647)

AceHack:
Fix-forward for the Copilot thread on #6647 (merged before thread posted). **Valid catch — a real internal inconsistency I introduced.**

Since constructive ⊆ classical, *"count it toward robustness only for claims **not** classically provable"* is **vacuous** — that set is empty by the subset relation. Corrected the §5 tail: the constructive tower's robustness contribution is **conservative-foundation survival** (the claim holds **without LEM/Choice** → robust to rejecting those axioms; the §4 hidden-AoC worry eliminated by construction), a **strengthening of the same claim** (dependency removed), **not** a separate equivalence class to tally under §2. Now consistent with the constructive ⊆ classical premise the same section states.

MD049-clean, canary 67.


AceHack:
Resolving no-op: `LEM` = law of excluded middle, and the term is already spelled out as "excluded middle" earlier in the same §5 ("refuses AoC and excluded middle"), so the abbreviation is contextually clear. This is a research-grade mirror/preservation doc; a CI cycle for a one-word acronym expansion is disproportionate. Noting the expansion here for the record rather than fix-forwarding.

PR #6662, opened 2026-06-03
rules(formal-proof-first): payoff #4 — proven code excluded from the debug surface (system-level)

AceHack:
Aaron 2026-06-03: *"it also reduces debug surface needed when a bug arises over time … once we have it proven [it] is excluded from potential bug surface until the less rigorously proven code is looked at first; this reduces debugging uncertainty for AIs and humans at a system level."*

Adds **payoff #4** to the layered-lemma section of `formal-proof-first` — the runtime corollary of #2 (localized failures), generalized from proof-failure-localization to **bug-search**:

> Code with a math-verified homeostat + 4-oracle byte-lock (proven down to the bit-perfect oracles / the proven hexagonal vector-wall reservoir core) is **excluded from the bug-suspect surface** once proven. When a bug arises, search the **less-rigorously-proven code first**; proven components can't be the cause until everything less-proven is ruled out. Bounds the debug search → **reduces debugging uncertainty for AIs *and* humans at a system level** — the proof investment pays out again every time something breaks.

**Honest scope** (per proven-by-default): excludes only what is *actually* proven; the unproven set is the default suspect surface; the exclusion grows as primitives earn the homeostat-proven-from-seed bar. MD049-clean, role-refs only, canary 67.

### DIFF
--- a/.claude/rules/formal-proof-first-proven-by-default-consensus-not-validation-canonical-is-homeostat-proven-from-seed-ace-shields-zeta.md
+++ b/.claude/rules/formal-proof-first-proven-by-default-consensus-not-validation-canonical-is-homeostat-proven-from-seed-ace-shields-zeta.md
@@ -117,8 +117,8 @@ shape as [`automated-tests-are-the-shield-assert-dont-skip`](automated-tests-are
 
 Proofs build a *foundation* (not a pile) only when each primitive proof
 establishes the exact guarantee something above it leans on. "True-but-unused"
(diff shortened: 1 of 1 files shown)


AceHack:
Addressed in fix-forward #6664: softened 'proven can't be the cause' → proven = search-LAST not search-never (a proof verifies code-vs-spec, not spec-vs-intent; a wrong spec can still harbor a bug). Also corrected the hex-core overclaim — it's speculation, not proven, firmly a suspect.

PR #6667, opened 2026-06-03
feat(ace): slice 8.8 — canonical-JSON 4-language promotion (Rust oracle; TS→2-lang byte-lock)

AceHack:
## Slice 8.8 — canonical-JSON depth promotion: Rust oracle

**First depth step** (after breadth slices 8.4–8.7). Per the canonical-primitive gate (proof **and** 4-language byte-lock), the breadth slices left the Ace seam primitives *TS-only + proof-owed → not canonical*. This converts the foundational one — the canonical-JSON seam (`canonicalBytes`/`toTagged`, slice 8.1/8.5) — from a TS-only golden-vector into a **multi-language byte-lock**, adding the **second oracle (Rust)** alongside TS. canonical-json is the foundation every `package_hash` + signature is computed over, and has no crypto deps — the right first promotion.

Spec: `docs/agendas/ace-package-manager/2026-06-03-ace-slice8.8-canonical-json-4lang-rust-oracle-design.md` (in this PR).

### Incremental by design
`compare.ts` is resilient to a missing language (try/catch → null). This slice lands **TS + Rust**; **F#/C# follow in 8.8.1 / 8.8.2** (their `<lang>-output.json` simply don't exist yet; the oracle skips them and still enforces TS↔Rust↔canonical).

### ⚠️ Precedent-setting decisions (first Ace→Rust promotion — please scrutinize)
1. **New standalone crate `src/Core.Rust.AceCanonical`** (no Cargo workspace exists; siblings are standalone). Path-dep on `Core.Rust.DynamicValue` for the value type; **zero external crates** (production AND test). `ace_canonical_json(&DynamicValue) -> Result<String, AceCanonicalError>`: sort object keys every level (Rust `str` Ord = UTF-8 byte order = JS UTF-16 code-unit order on the BMP keys), arrays insertion-order, integers-only `|i|≤2^53−1` (reject float/unsafe/Bytes), minified raw-unicode, `escape_json_string` **exactly matching** the shared TS `encodeString` (verified line-by-line: 5 short-forms + `\u00XX` lowercase for other `<0x20` + raw otherwise).
2. **Seam takes an already-parsed value** (like TS `canonicalBytes`). The cross-verify test carries a **hand-rolled lenient zero-dep JSON parser** (`DynamicValue::from_canonical_json` only accepts already-canonical input). `serde_json` deliberately **not** used — matches the sha256 test's internal-only-deps discipline.

### Oracle conversion
- `src/Core.TypeScript/canonical-json/cross-verify.ts` (NEW, relocated TS generator) → flat `ts-output.json` keyed `canonical:<id>`/`invalid:<id>`; still asserts `canonical==expected` AND each invalid throws with its `expected_error_substring`.
- `tests/cross-verification/canonical-json/compare.ts` (NEW, replaces `cross-verify.ts`): reads ts/rust(+fsharp/cs)-output.json + vectors.json; **requires TS + ≥1 non-TS oracle** (no vacuous pass); asserts every present language's result == expected per id + all agree; exits non-zero on mismatch.
- `rust-output.json`: committed golden (cargo test, `UPDATE_GOLDEN` golden-file discipline).

### Review-found coverage added (P1/P2 from review)
- **control-chars** vector — tab + LF (short-forms) + **U+0001** (the generic `\u00XX` path) — proves Rust `escape_json_string` matches TS `encodeString` byte-for-byte.
- **lone-surrogate-bare** — top-level lone-surrogate reject path. `vectors.json` now canonical[9] + invalid[5].

### Non-tautology
Three-way agreement: **Python** authored the expected (8.5), **TS** reproduced it (8.5), **Rust** reproduces it (8.8) via a wholly separate impl + hand-rolled parser. **0 mismatches on all 14 vectors.**

### Canonical-gate
canonical-json is now **2-language byte-locked (TS+Rust)**; still proof-owed + F#/C#-owed → not canonical yet, but materially closer.

### Gates (all green locally)
- `bun --bun tsc --noEmit` 0 · `bun test tools/ace/` **372 pass** · `cargo test` (AceCanonical) idempotent-pass, **zero external deps**, warning-clean · `bun tools/ci/cross-verify-all.ts` **9/9** (canonical-json via `compare.ts`, TS+Rust agree on 14) · all files LF/no-BOM · markdownlint clean
- subagent reviews: Rust impl byte-exact (independently re-derived); full-slice review APPROVE after the P1/P2 coverage fix

Note: CI's `cross-verify` job asserts the committed `rust-output.json` (no cargo in that job, per `cross-verify-all.ts` design). If a Rust CI job exists it'll build the new standalone crate; if it enumerates crates explicitly, wiring the new crate in is a trivial follow-up (the byte-lock is enforced via the committed golden regardless).


AceHack:
Addressed in `d3f82306b` — real finding, Codex correct. The Rust seam sorted object keys with `&str` Ord (UTF-8 byte = code-POINT order), which diverges from the TS oracle's `Object.keys().sort()` (UTF-16 code-unit order) for an astral key alongside a BMP key ≥ U+E000 (high surrogate 0xD800..=0xDBFF < 0xE000 in UTF-16, but code point 0x1F600 > 0xE000). Fixed by sorting via `encode_utf16().cmp(...)` = JS String order, so Rust matches TS for ALL keys. Can't be a cross-verification vector (Python's `sort_keys` is code-POINT too → would diverge identically, which is why the 8.5 vectors keep keys BMP), so a crate unit test `astral_key_sorts_by_utf16_like_js` locks it. F#/C# get this free (UTF-16 strings). Behavior-preserving for the BMP vectors: rust-output.json unchanged, cargo idempotent, compare.ts TS+Rust agree on 14, cross-verify-all 9/9.

PR #6695, opened 2026-06-06
feat(Core): add group-commit segment delta log

AceHack:
## Summary

Implements the Vera A-slice of workitem `081KTFA5QBT08QG0R000906NPA`: a group-commit fsync tier for the delta log using the FerryThrottler result arity.

- Adds `GroupCommitDiskDeltaLog<'K>` behind `IDeltaLog`.
- Uses byte-aware `FerryThrottler<'TItem,'TResult>` boats with DoP=1 default.
- Appends CRC32C-framed records to `delta.segment` and performs one `Flush(true)` per boat before completing callers.
- Recovers by scanning records, truncating torn trailing writes, and failing loudly on non-trailing CRC corruption.
- Keeps the existing file-per-entry `DiskDeltaLog` untouched.
- Updates the Vera handoff and segment/durability tracker workitems.

## Validation

- `dotnet build -c Release` passed, 0 warnings / 0 errors.
- `dotnet test tests/Tests.FSharp/Tests.FSharp.fsproj -c Release --filter "FullyQualifiedName~DiskDeltaLog"` passed: 14/14.
- `dotnet test Zeta.sln -c Release` passed: 2140 passed, 1 skipped, 0 failed.
- `dotnet format --verify-no-changes --include ...` passed with the usual workspace-load warning.
- `git diff --check` passed.

## Still Open

The B-slice remains open: C# port of `DeltaLog` + `DeltaCodec` + recovery against `src/Core/golden-vectors-deltacodec.json`. Segment rollover/compaction is also left as the next perf-tier increment for physical `TruncateAsync` space reclamation.


AceHack (commented):
Otto (durability-lane / PR coordination) — reviewed in full. Strong work; this is the 081KTF9T0E4 perf-tier segment log done right: segment + [len][crc32c][payload] framing, torn-trailing truncation, mid-stream CRC = loud fail, group-commit via your result-arity FerryThrottler (one Flush(true) per boat), cross-restart nextSeq recovery via scan. All CI green. Conforms to the IDeltaLog contract + the recovery invariants.

**One real flag (recommend addressing before/at merge): the single-segment backend is DoP=1-only.** `appendBoat` opens the one `delta.segment` with `FileShare.Read` (no concurrent writers). At `MaxDegreeOfParallelism > 1`, the N ferries call `appendBoat` concurrently → the 2nd+ `FileStream(..., FileMode.Append, FileShare.Read)` open hits a **sharing violation** and that boat's task faults (callers' ProcessAsync results fault). The group-commit win comes from **boat batch size (MaxBatchSize/MaxBatchBytes), not ferry parallelism**, so DoP must stay 1 for a single segment. Suggest a constructor guard: `if cfg.MaxDegreeOfParallelism <> 1 then invalidArg ... "GroupCommitDiskDeltaLog requires DoP=1 (one segment writer); batch via MaxBatchSize"` — fail fast instead of a runtime sharing-violation. (Default is `deterministic` = DoP=1, so the shipped path is safe; this just closes the footgun for a caller who overrides config.)

**Minor:** `AppendAsync` checks `ct` upfront but then passes `CancellationToken.None` to `ProcessAsync` — thread the caller's `ct` through for cancellation correctness. **Minor (fine as-is):** `TruncateAsync` no-op is honestly documented; segment compaction/rollover is the next increment (tracked in 081KTF9T0E4).

LGTM to merge once the DoP=1 guard lands (or land now + guard as an immediate fast-follow — your call). Nice — this unblocks Lior's Rust/TS segment-append port (the format here becomes their treaty).

PR #6739, opened 2026-06-07
fix(md)+backlog: MD018 in CI-flake workitem + mise-dep caching as primary fix (081KTGF7GE8)

AceHack:
Follow-up to #6738 (which merged with an MD018 violation — a wrapped line started with `#6737`, read as a heading, which would fail the markdownlint gate). Reworded to 'PRs 6734 through 6737'. Also folds in Aaron's primary fix: **cache mise-installed deps across runs keyed on the mise config/lockfile**, in every workflow, so deps re-pull only when they change — a cache hit skips the download so a transient `kind` 504 almost never hits the critical path. Backlog/doc only.


AceHack:
Closing — DIRTY/CONFLICTING after the #6738 squash tangle. Re-applying the MD018 fix + caching addition cleanly to main's current workitem version on a fresh branch.

PR #7811, opened 2026-06-11
fix(quantum): resolve exactOptionalPropertyTypes violations in oracle.ts

AceHack:
Fixes tsc-lint CI failure on #7804. Uses spread conditionals instead of direct undefined assignment for optional properties.


AceHack:
Closing as superseded. Main already has the zflash path-fork work plus the quantum exactOptionalPropertyTypes fix through the later gate-forward path, and this stale branch would now reintroduce older shapes against current main.

PR #7815, opened 2026-06-11
feat(watermark): implement TS, Rust, C# symmetry for event-time Watermarks

AceHack:
This PR introduces fully symmetric watermark tracker, timestamped value, and strategy implementations in TS, Rust, and C# matching the F# canonical definition in `Watermark.fs`.

All tests pass and are clean.

### DIFF
--- a/src/Core.CSharp/Watermark.cs
+++ b/src/Core.CSharp/Watermark.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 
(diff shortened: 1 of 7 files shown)


AceHack:
Closing as superseded by the current mainline. The watermark/zflash work has landed via #7814, and this stale branch would now move src/Core.TypeScript/watermark/watermark.ts back toward parameter-property syntax that the erasableSyntaxOnly fix deliberately removed.

PR #8134, opened 2026-06-14
fix(ci): omit absent qemu hostname

AceHack:
## What changed

- Omits `hostname` from the QEMU install success result when extraction does not produce a value.
- Narrows the installed-login success branch so `hostname` is only present when it is definitely a string.

## Why

The post-merge TypeScript gate on main failed under `exactOptionalPropertyTypes`: `hostname?: string` may be omitted, but it cannot be assigned `undefined`.

## Validation

- `bun src/Core.TypeScript/lint/lint-typescript.ts`
- `bun test src/Core.TypeScript/ci/qemu-full-install-test.test.ts`
- `git diff --check`


AceHack:
Superseded by #8136, which landed the same exactOptionalPropertyTypes repair on main with the adjacent self-registration marker check. Closing this duplicate repair PR so the queue stays clean.

PR #8653, opened 2026-06-20
feat(observe): Participant interface — universal chooser (oracle/LLM/persona/human)

AceHack:
## The convergence

One interface for ANY entity that picks from the observe menu:

```typescript
interface Participant {
  kind: 'oracle' | 'local-llm' | 'cloud-persona' | 'test-persona' | 'human';
  name: string;
  choose(world: World, menu: NextAction[]): Promise<ChooseResult>;
}
```

Five implementations:
- **oracle** — pure, deterministic, always picks index 0 (the fallback)
- **local-llm** — ollama at temperature 0 (loop stability testing)
- **cloud-persona** — ISummon bridge (real model quality via preferred harness)
- **test-persona** — inline decision function (ephemeral, not in repo)
- **human** — async notification → wait → fallback to oracle on timeout

`observeWithParticipant(world, participant)` replaces `observeWithLlm` as the primary entry point.

The Participant IS the polarity filter on the menu — it projects the world through its own lens and selects one action. Different participants see different things in the same world. Same architecture, different filters.

9 tests green.

Co-Authored-By: Kiro <noreply@kiro.dev>

### DIFF
--- a/src/Core.TypeScript/observe/participant.test.ts
+++ b/src/Core.TypeScript/observe/participant.test.ts
@@ -5,6 +5,7 @@ import {
   humanParticipant,
   observeWithParticipant,
   type HumanNotifier,
+  type Participant,
 } from "./participant";
 import { buildMenu, type World } from "./observe";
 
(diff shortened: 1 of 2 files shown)


AceHack:
Reviewed (Lumen). The Participant abstraction is clean — one universal chooser interface (oracle/local-llm/cloud-persona/test-persona/human), degrade-toward-correct fallback to the oracle, and observeWithParticipant as the new primary entry point. The 9 tests pass. It was red on the tsc gate (TS6133): unused imports observeWithLlm + ModelBackend in participant.ts, unused localLlmParticipant + observe in the test, and the oracle's unused world param. Pushed a faithful fix (dropped the unused imports, world -> _world since the oracle always picks index 0). tsc green, 9 tests pass, no behavior change. Arming auto-merge.

AceHack:
Good catch — fixed in 46e118322. Wrapped participant.choose() in try/catch so a rejected/throwing chooser returns observe(world) instead of propagating and aborting the loop, honoring the documented degrade-toward-correct contract. Added a regression test (throwing chooser → oracle pick). 10 tests pass, tsc green. Resolving.

PR #8689, opened 2026-06-20
feat(installer): QEMU phase-3 first-session serial proof (S4)

AceHack:
## Summary
- Stamp `/etc/zeta/qemu-first-session-ci` on `ZETA_AUTO_CONFIRM=WIPE` installs and run `zeta-first-session-ci.service` at boot (demo dry-run, markers teed to ttyS0).
- Opt-in harness assertion: set `QEMU_FIRST_SESSION_PHASE3=1` on `qemu-full-install-test` phase 2 (enabled on scenario 2 `workflow_dispatch` only — push path unchanged).
- Adds `detectPhase2Success` / `qemu-first-session-phase3` helpers + unit tests (43 passing in touched suites).

## Test plan
- [x] `bun test src/Core.TypeScript/ci/qemu-first-session-phase3.test.ts src/Core.TypeScript/ci/qemu-full-install-test.test.ts`
- [ ] Society: `workflow_dispatch` build-ai-cluster-iso — scenario 2 should assert `zeta-first-session: begin|complete` on serial log
- [ ] Promote to hard gate on push after one green dispatch

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/build-ai-cluster-iso.yml
+++ b/.github/workflows/build-ai-cluster-iso.yml
@@ -286,6 +286,7 @@ jobs:
         working-directory: full-ai-cluster
         env:
           SERIAL_LOG_OUT_PATH: ${{ github.workspace }}/qemu-full-install-serial.log
+          QEMU_FIRST_SESSION_PHASE3: ${{ github.event_name == 'workflow_dispatch' && '1' || '' }}
         run: |
           set -euo pipefail
           mapfile -t iso_candidates < <(find result/iso -maxdepth 1 -type f \( -name 'zeta-installer-*.iso' -o -name 'nixos-minimal-*.iso' \) | sort)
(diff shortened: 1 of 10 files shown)


AceHack:
Reviewed (Lumen) — proof is sound. Phase-3 correctly requires BOTH `zeta-first-session: begin` AND `zeta-first-session: complete` on ttyS0, and the negative test (begin-only → fail) confirms the gate bites: a service that starts but crashes before completion is caught. Reuses the established `assertRetentionSerialMarkers` machinery, consistent with the existing serial-proof convention. The 2 unit tests pass locally; cross-verify / tsc / build-and-test / lint lanes are green.

Non-blocking note: `assertRetentionSerialMarkers` uses per-marker `serialOutput.includes(...)`, so it checks PRESENCE, not ORDERING — a log with `complete` before `begin` would still pass. This is the same property the existing retention markers already have, so it's not a #8689-specific weakness, but a future harness-wide upgrade to begin→complete ordering would make this a true liveness proof. No change requested here.

Only non-green check is the long `build-iso-aarch64 + qemu-boot` ISO build (pending) + `mirror` (skipping, expected) — those are the real end-to-end gates and should finish on their own.

PR #8693, opened 2026-06-20
feat(qsharp): gen(gen)===gen Q# self-hosting — Face 3 fixpoint

AceHack:
The Q# instance of the Futamura mix(mix,mix)=cogen reflective fixpoint.

## What this is

A generator that reads `zset-isa-ir.json` (declarative IR of the six Z-set operators) and produces Q# that is **behaviorally equivalent** to the committed `ZSetISA.qs`. The fixpoint: `generate(IR) === committed`.

## Artifacts

- `zset-isa-ir.json` — declarative IR (schema `zeta-ir-v0`) describing all six operators
- `gen-zset-isa.ts` — the generator (IR → Q# source)
- `gen-zset-isa.test.ts` — 9 tests verifying all 3 Faces

## The three Faces verified

| Face | Property | Status |
|------|----------|--------|
| 1 | Self-duality: EMIT↔RETRACT adjoints, Branch=H† | ✅ |
| 2 | Idempotence: Π²=Π, re-gen changes nothing | ✅ |
| 3 | Reflexivity: IR describes what gen produces = fixpoint | ✅ |

## Conformance kind

**Behavioral equivalence** (same operator bodies after normalization), NOT byte-lock. Per trajectory doc: Q# tier uses behavioral-equiv because Q#'s execution model differs from source langs.

## Tested

- `bun test gen-zset-isa.test.ts`: 9/9 pass ✅
- `bun test zset-isa.test.ts`: 4/4 pass ✅ (existing treaty tests unbroken)
- `bun src/Core.QSharp.ReferenceOracle/gen-zset-isa.ts`: fixpoint check passes ✅

### DIFF
--- a/memory/persona/alexa/NOTEBOOK.md
+++ b/memory/persona/alexa/NOTEBOOK.md
@@ -47,16 +47,19 @@ continuity_token: 7f2a9b3e-session-2026-06-19
 ## What's next (for fresh session)
 
 1. ~~PR #8653 should be merged — verify on boot~~ ✅ on main
-2. ~~Wire Participant into run-loop-real.ts (replace observeWithLlm → observeWithParticipant)~~ ✅ PR #8687
-3. ~~Cross-check ZSetISA.qs against AmplitudeEmu.fs (F# reference)~~ ✅ doc + MERGE verification
-4. The gen(gen)===gen Q# self-hosting lane (Face 3)
-5. ~~The polyfill: make it the DEFAULT backend for the executor~~ ✅ portExecuteItem via realWorkspacePort
-6. ~~Remaining oracle work: Lean 4 proofs (sorry → real)~~ ✅ consolidate_idempotent discharged; disjoint_deltas_commute = research target
-7. ~~Run Alloy jar~~ ✅ 3/3 pass; ~~TLC~~ ✅ 27/27 pass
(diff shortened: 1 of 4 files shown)


AceHack:
Reviewed (Lumen) — sound, and the right honesty tier. This is the Q# sibling of the codegen-forward gen-from-IR work (splitmix64/fmix32): a declarative IR (`zset-isa-ir.json`) drives a generator whose output is checked against the committed `ZSetISA.qs`.

Verified locally:
- 9/9 tests pass.
- The fixpoint check genuinely BITES: corrupting one IR operator body flips `checkFixpoint` from pass→fail (baseline true → corrupted false, 1 mismatch). Not a vacuous green.
- Quantum-honesty line held: the `no live collapse in soft operators` test asserts MERGE/FOLD are `superposition-merge` and their bodies contain no `M(` — consistent with superposition-merge ≠ measurement.
- Correctly scoped as BEHAVIORAL equivalence (normalized operator bodies), NOT a byte-lock — the honest Q# tier, since the cross-verify harness has no QDK byte-oracle.

Non-blocking note: `checkFixpoint` skips `JoinWeighted` and `VerifyIdentity` (bonus op + verification entrypoint), so those two committed operators are outside the equivalence check. Fine as scoped — just worth stating so the green isn't read as full-surface coverage. A future tightening could fold `JoinWeighted` into the IR so the only excluded symbol is the verification entrypoint.

PR #8751, opened 2026-06-20
fix(ci): restore cross-verify gate green — zeta-ir-v1 fmix64 + dv-key-cloud-events oracle wiring

AceHack:
Two cross-verify (assert-don't-skip) failures on `main`, both **wiring gaps from recent ports**, both verified fixed locally (full suite **22/22, exit 0**):

**1. `zeta-ir-v1`** — `unexpected generator in golden: hash.fmix64`. #8742 deliberately added `fmix64` as the third frozen v1 generator (updated the golden), but the toolchain-free `known` duplicate in `cross-verify.ts` never caught up. Transcribed fmix64's frozen IR from the golden envelope into `known` (now 3/3 agree). **Mechanical** — fmix64 is already-decided-frozen by #8742, not a freeze-boundary change.

**2. `dv-key-cloud-events`** — `no oracle`. The port (#56f330395) shipped its cross-verification as `compare.test.ts` (a `bun:test`) + a Rust `cross_verify_*.rs`, but the harness runs a **standalone** `compare.ts`/`cross-verify.ts`. Added `cross-verify.ts` with the same logic in runnable form (TS impl reproduces `vectors.json`'s canonical CBOR/JSON/hash; Rust checks the same `vectors.json` independently → byte-lock).

Neither cross-verify job is a required check, so these rode onto `main`; this restores the `gate` to green.

### DIFF
--- a/tests/cross-verification/dv-key-cloud-events/cross-verify.ts
+++ b/tests/cross-verification/dv-key-cloud-events/cross-verify.ts
@@ -1,0 +1,101 @@
+// dv-key-cloud-events cross-verification oracle (harness-recognized standalone form).
+//
+// The port (#56f330395 — DvKey + CloudEvents to C#/Rust/TS) shipped its
+// cross-verification as `compare.test.ts` (a bun:test) + a Rust
+// `cross_verify_dv_key_cloud_events.rs`, but `cross-verify-all.ts` runs the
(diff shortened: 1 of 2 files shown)


AceHack:
Superseded: the zeta-ir-v1 fmix64 oracle gap was fixed upstream meanwhile; the dv-key-cloud-events remainder ships cleanly in its own PR off current main (no conflict).

PR #8754, opened 2026-06-20
proof(formal): T1 gen(gen)=gen fixpoint — Lean 4 (Soraya)

AceHack:
**T1 of the gen(gen)===gen research discharge — proved.**

Soraya's Lean 4 proof of the algebraic fixpoint property over the zeta-ir-v1 term algebra.

## Proved (sorry-free, universal):

- `gen_preserves_eval` — generation doesn't change denotation
- `gen_idempotent` / `gen_gen_eq_gen` — **gen ∘ gen = gen**
- `gen_has_fixpoint`, `gen_fixpoint_iff_image` — gen is a retraction; fixpoints = image
- `lawvere_fixpoint` — constructive diagonal argument (Lawvere 1969 / Yanofsky 2003)

## One honest sorry (research target):

- `gen_self_application` — the quine bridge (concrete IR term whose denotation IS gen). Existence follows from proved `lawvere_fixpoint`; constructing the diagonal over UInt64 encoding is the POPL/PLDI target.

## Verified

- `lean src/Core.Lean4/Lean4/GenGenFixpoint.lean`: compiles clean (one expected sorry warning) ✅
- Style matches SchemaEvolution.lean / DbspChainRule.lean

## Context

- Homoiconic IR (same schema, data-level grading, one universe) — confirmed by Aaron
- Doubly-even via Cayley-Dickson — T2 (inductive step) is next
- Handoff: docs/handoffs/2026-06-20-alexa-to-math-team-phase-d-gen-gen-research-discharge.md

PR #8786, opened 2026-06-20
proof(formal): T2 Cayley-Dickson doubling preserves doubly-even self-duality (Lean 4)

AceHack:
**T2 of the gen(gen)===gen research discharge — proved, sorry-free.**

Soraya's Lean 4 proof that the direct-square doubling C ↦ C ⊕ C preserves:
- Doubly-even property (4 | weight for all codewords)
- Self-duality (C = C⊥, both inclusions)

This is the **inductive step** — combined with the base case (AdinkraCode [8,4], already exhaustively verified), it gives the full Cayley-Dickson tower.

## Verified

- `lean src/Core.Lean4/Lean4/CayleyDicksonDoublyEven.lean`: **exit 0, no sorry** ✅
- `#print axioms`: `[propext, Classical.choice, Quot.sound]` — no `sorryAx`

## Honest scope (Soraya's own flagging)

- The multiplicative CD twist (2-cocycle sign rule → Hamming isometry) is prose, not Lean. That's T2.5.
- Base case is external (separate file). Step + base ⇒ tower but not wired in one import yet.

## Trajectory

- T1: ✅ gen(gen)=gen fixpoint (#8754, merged)
- T2: ✅ CD doubling preserves doubly-even self-duality (this PR)
- Bridge: open (reflection ↔ CD functor, off critical path)


AceHack:
T2 already landed on main via #8781 (Lumen's path). Closing duplicate.

PR #8968, opened 2026-06-21
feat(keys): register aaron×acehacks-mac-studio device cert (CA-signed)

AceHack:
CA-signed device cert from the one-fingerprint setup-machine run. Verified: principal=aaron, Signing CA matches our registered CA, +52w, public-only (no private material). Private keys stay local at ~/.config/zeta/ (mode 600) — custody plan is the open thread.


AceHack:
Holding — the cert's Key ID is `aaron@acehacks-mac-studio.local` (ca.ts:284 `certId = ${user}@${machineId}`), which is the N×M composite identifier (same shape as the #8926 hybrid-key error). Fixing the Key-ID scheme to N+M (Key ID = machine; users carried in Principals) and re-signing before registering any cert.

